PR #20, opened 2024-09-24
add initial version of ΔQ tooling in Rust

rkuhn:
This commit adds the initial version of a ΔQ tool written in Rust. Please see the README.md for a list of shortcomings (a.k.a. things to work on next). The intention is to eventually put something like this on https://leios.cardano-scaling.org/, in which case we’ll probably want to move the `dq.eval(&ctx)` to the client (in a web worker, e.g. using a yew agent). It remains to be decided how we foster collaboration on ΔQ models, i.e. how to share and update the expressions between multiple people.

In case it isn’t glaringly obvious: while the CSS makes the web UI somewhat workable, it is by no means to be considered a design :-) I’m very much open for any kind of help in this regard!


rkuhn:
Ah, one more thing: I didn’t check in `.github/workflows` because I haven’t yet checked how this repository is supposed to work. You can see my prior setup at https://github.com/rkuhn/deltaq-rs/blob/master/.github/workflows/rust.yml

bwbush:
> Ah, one more thing: I didn’t check in `.github/workflows` because I haven’t yet checked how this repository is supposed to work. 

I'll add this to the Nix flake and, based on that, to the CI workflow. I created #23 to track this.

Please go ahead and merge this PR.

abailly-iohk:
@hajola Might be able to help us on the design, he did a great job on https://peras-simulation.cardano-scaling.org/index.html

PR #28, opened 2024-09-27
Bunch of old patches

dcoutts:
These are patches I had locally from a couple years ago that never made it to the upstream repo.

Some rebasing & edits may be needed to make them work in the current code.


Saizan:
```
By default, secrets are not passed to workflows triggered from forks, including Dependabot
```
@wenkokke seems like PRs from forks won't be able to pass checks because they will lack credentials.

PR #36, opened 2024-10-03
Rust simulator L1 scaffolding

### LINKED ISSUE #34: Rust simulation of L1 chain
Build a basic simulation of the L1 chain, which we can simulate Leios on top of.

- Accept input from a file (ideally match the format used by the haskell node when relevant)
- Emit a stream of events as output
- Simulate a VRF slot lottery to decide which node(s) publish
- Do not simulate transactions in detail, initially we just care about block throughput

SupernaviX:
Scaffolding for (a command-line version of) the L1 chain for the rust Leios simulator. This is very WIP (I'm still trying to understand a lot about the problem, including what input/output should even look like), but I'm opening a PR in the interest of transparency.

This isn't a long-lived branch, I expect to merge (or discard) these changes after a week or two.

Fixes #34 

So far, this has a fake-VRF lottery for block production, and logs stats about how many blocks were published and by which pools.


sim-rs/src/sim.rs
```diff
@@ -0,0 +1,90 @@
...
+
+pub struct Simulation {
+    pools: Vec<Pool>,
+    vrf_max_score: u64,
+    current_slot: u64,
+}
+
+impl Simulation {
+    pub fn new(config: SimConfiguration) -> Self {
+        let mut rng = ChaChaRng::from_rng(thread_rng()).expect("couldn't initialize RNG");
```
Quantumplation, line 23: probably want the seed as part of the SimConfiguration

SupernaviX: @Quantumplation I don't want to make guarantees of determinism unless it's necessary. The `ce_netsim` library forces everything to be "realtime", and I'm worried that determinism will be hard to maintain if actual threads are interacting.

Quantumplation: Even outside of making it deterministic, just generating things like the same slot schedule, connections, topology, etc. over multiple runs is super useful.

abailly-iohk:
> This isn't a long-lived branch, I expect to merge (or discard) these changes after a week or two.
😂 To me a short-lived branch is one that is not older than 2 days

SupernaviX:
> > This isn't a long-lived branch, I expect to merge (or discard) these changes after a week or two.
> > 😂 To me a short-lived branch is one that is not older than 2 days

I merged it, but developing against main on a shared repo feels icky

PR #40, opened 2024-10-08
Provide a `list` command to list available visualisations

abailly-iohk:
Also provide a `run` command to, well, run the simulation and rename some types to use less abbreviations.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -14,5 +14,8 @@ index-state:
   , hackage.haskell.org 2024-03-21T15:07:04Z
   , cardano-haskell-packages 2024-03-21T19:04:02Z
 
-packages: 
+tests: True
+test-show-details: direct
+
+packages:
   simulation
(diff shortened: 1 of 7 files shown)


wenkokke (approved):
Looks good. Oddly, I already formatted the code with fourmolu in an earlier commit. I should investigate whether or not my formatter is picking up on the configuration.

abailly-iohk:
I have been bitten in the past by different versions of fourmolu. I am not using nix so perhaps it's _my_ version that's misaligned with everyone else's?

wenkokke:
> I have been bitten in the past by different versions of fourmolu. I am not using nix so perhaps it's _my_ version that's misaligned with everyone else's?

I'm simply using the latest version, so if we're meant to use an older version, then it's likely I am off.

abailly-iohk:
```
% fourmolu --version
fourmolu 0.14.0.0
using ghc-lib-parser 9.6.3.20231014
```

PR #63, opened 2024-11-01
Formal specification: Base chain

yveshauser:
__Base chain__ rules from Figure 2: The Simplified Leios Protocol

### DIFF
--- a/formal-spec/Leios/SimpleSpec.agda
+++ b/formal-spec/Leios/SimpleSpec.agda
@@ -129,7 +129,10 @@ data _⇀⟦_⟧_ : Maybe LeiosState → LeiosInput → LeiosState × LeiosOutpu
        ────────────────────────────────────────────────────────────────────────────────────
          just s ⇀⟦ SLOT ⟧ (s , EMPTY)
 
-  Base₂ : ∀ {bs bs' txs} → let open LeiosState s in
+  Base₂ : ∀ {bs bs'} →
+         let open LeiosState s
+             txs = MemPool
+         in
        ∙ ∅ˢ ≡ filterˢ (λ eb → isVote2Certified s eb × eb ∈ᴮ slice L slot 2) (fromList EBs)
        ∙ bs BF.⇀⟦ B.SUBMIT (inj₂ txs) ⟧ (bs' , nothing)
        ────────────────────────────────────────────────────────────────────────────────────
--- a/formal-spec/Leios/SimpleSpec.agda
+++ b/formal-spec/Leios/SimpleSpec.agda
@@ -33,6 +33,7 @@ postulate VTy FTCHTy : Type
 
 data LeiosInput : Type where
   INIT     : VTy → LeiosInput
+  SUBMIT   : List Tx → LeiosInput
   SLOT     : LeiosInput
   FTCH-LDG : LeiosInput
 
@@ -102,8 +103,8 @@ data _⇀⟦_⟧_ : Maybe LeiosState → LeiosInput → LeiosState × LeiosOutpu
 
   Init : ∀ {V bs bs' SD} →
        ∙ {!!} -- create & register the IB/EB lottery and voting keys with key reg
-       ∙ bs BF.⇀⟦ B.INIT {!V_chkCerts!} ⟧ (bs' , just (B.STAKE SD))
-       ─────────────────────────────────────────────────────────────
+       ∙ bs BF.⇀⟦ B.INIT {!V_chkCerts!} ⟧ (bs' , B.STAKE SD)
+       ──────────────────────────────────────────────────────
        nothing ⇀⟦ INIT V ⟧ (initLeiosState V SD , EMPTY)
 
   -- Network and Ledger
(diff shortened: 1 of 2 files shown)


formal-spec/Leios/SimpleSpec.agda
```diff
@@ -115,45 +121,54 @@ data _⇀⟦_⟧_ : Maybe LeiosState → LeiosInput → LeiosState × LeiosOutpu
        ────────────────────────────────────
        just s ⇀⟦ FTCH-LDG ⟧ (s , FTCH-LDG l)
 
-  -- TODO: Base chain
+  -- Base chain
+
+  Base₁ : ∀ {bs bs' eb txs} → let open LeiosState s in
```
WhatisRT, line 126: I think steps 1 and 2 should be separate. Storing transactions happens if `SUBMIT` arrives, but submitting an EB/transactions to the base chain happens once per slot (which is actually quite confusing to me, I think it should only happen once per slice maybe?). So both of these should be able to happen without the other.

formal-spec/Leios/SimpleSpec.agda
```diff
@@ -115,45 +121,54 @@ data _⇀⟦_⟧_ : Maybe LeiosState → LeiosInput → LeiosState × LeiosOutpu
...
+       ∙ eb ∈ filterˢ (λ eb → isVote2Certified s eb × eb ∈ᴮ slice L slot 2) (fromList EBs)
+       ∙ bs BF.⇀⟦ B.SUBMIT (inj₁ eb) ⟧ (bs' , B.EMPTY)
```
WhatisRT, line 128: Giorgos confirmed that the base chain ignores this if it doesn't make a block in the current slot. I think we should make at least a note of this somewhere.

yveshauser: How about adding a predicate `canProduceRB` that checks that the ranking block can be added by the party?

formal-spec/Leios/SimpleSpec.agda
```diff
@@ -115,45 +121,62 @@ data _⇀⟦_⟧_ : Maybe LeiosState → LeiosInput → LeiosState × LeiosOutpu
...
-  -- TODO: Base chain
+  -- Base chain
+  --
+  -- Note: Submitted data to the base chain is only taken into account
+  --       if the party submitting is the block producer on the base chain
+  --       for the given slot
+
+  Base₁ : ∀ {txs} → let open LeiosState s in
+        ──────────────────────────────────────────────────────────────────
+        just s ⇀⟦ SUBMIT (inj₂ txs) ⟧ (record s { MemPool = txs } , EMPTY)
```
WhatisRT, line 132: It's an interesting question what to do here. `SUBMIT` for the base chain has this kind of one time semantics and does not store anything, and doing it like this has a similar flavour. So for consistency this makes sense, but then we'd have to do a `SUBMIT` to Leios containing the most recent mempool every slot, which I'm not sure is what the paper intended. But now that I reread it, I think the paper is just unclear here, and it could be either option.

formal-spec/Leios/SimpleSpec.agda
```diff
@@ -115,45 +121,62 @@ data _⇀⟦_⟧_ : Maybe LeiosState → LeiosInput → LeiosState × LeiosOutpu
...
+
+  Base₂a : ∀ {bs bs' eb} → let open LeiosState s in
+         ∙ eb ∈ filterˢ (λ eb → isVote2Certified s eb × eb ∈ᴮ slice L slot 2) (fromList EBs)
+         ∙ bs BF.⇀⟦ B.SUBMIT (inj₁ eb) ⟧ (bs' , B.EMPTY)
+         ────────────────────────────────────────────────────────────────────────────────────
+         just s ⇀⟦ SUBMIT (inj₁ eb) ⟧ (s , EMPTY)
+
+  Base₂b : ∀ {bs bs'} → let open LeiosState s renaming (MemPool to txs) in
+         ∙ ∅ˢ ≡ filterˢ (λ eb → isVote2Certified s eb × eb ∈ᴮ slice L slot 2) (fromList EBs)
```
WhatisRT, line 141:
```suggestion
         ∙ ∅ ≡ filterˢ (λ eb → isVote2Certified s eb × eb ∈ᴮ slice L slot 2) (fromList EBs)
```
There's a `HasEmptySet` typeclass that provides this, so you don't need that `ˢ`. Also, using `≡` is fine here because we're comparing with the empty set, but in general you need to use `≡ᵉ` for the 'correct' set equality.

WhatisRT (approved):

formal-spec/Leios/SimpleSpec.agda
WhatisRT: Though I think if we keep this overwrite semantics we should probably rename `MemPool` because that's misleading. The only thing that comes to mind right now is `ToPropose`, which I think is also bad, but less so...

yveshauser: In _Figure 2: The Simplified Leios Protocol_ txs are the most recently stored transactions - therefore the overwrite semantics

WhatisRT: My initial understanding was that if it says 'let txs be the (most recently stored) transactions', it means that from the stored transactions you select a subset and then you'd have the append semantics. I think either interpretation is consistent with the paper, but this one is consistent with the base chain as well.

PR #65, opened 2024-11-05
Formal specification: ledger construction

yveshauser:


formal-spec/Leios/SimpleSpec.agda
```diff
@@ -52,6 +61,23 @@ record LeiosState : Type where
...
+    lookup IBs <$> i
+
+  lookupTxs : EndorserBlock → List Tx
+  lookupTxs = join ∘ map txs ∘ map body ∘ mapMaybe lookupIB ∘ join ∘ map ibRefs ∘ mapMaybe lookupEB ∘ ebRefs
+    where open EndorserBlockOSig
+          open IBBody
+          open InputBlock
+
+  constructLedger : List EndorserBlock → List Tx
+  constructLedger = join ∘ map lookupTxs
```
WhatisRT, line 79:
```suggestion
  lookupEB : EBRef → Maybe EndorserBlock
  lookupEB r = find (λ b → getEBRef b ≟ r) EBs

  lookupIB : IBRef → Maybe InputBlock
  lookupIB r = find (λ b → getIBRef b ≟ r) IBs

  lookupTxs : EndorserBlock → List Tx
  lookupTxs eb = do
    eb′ ← mapMaybe lookupEB $ ebRefs eb
    ib  ← mapMaybe lookupIB $ ibRefs eb′
    txs $ body ib
    where open EndorserBlockOSig
          open IBBody
          open InputBlock

  constructLedger : List EndorserBlock → List Tx
  constructLedger = concatMap lookupTxs
```
I found `lookupTxs` really hard to read, so I tried this. I'm not sure if having to know do notation and the List monad is better, but I think it's at least more readable in that case. If we don't want that, maybe some named helper functions might improve things?

PR #67, opened 2024-11-05
Replace hashing postulates with type classes

WhatisRT:
It took me 3 attempts with a whole bunch of complicated things until I realized that we can just do it like that. It's a little annoying that we now have to actually use instance resolution for types that are fully known, but I think it ticks all the boxes: 
- we can instantiate it however we like whenever we like (in particular we don't have to use GHC pragmas, which means we can offer a library and the user of the library can provide hashing capabilities)
- it doesn't clutter the spec beyond some paint threshold


yveshauser (approved):
Great!

PR #75, opened 2024-11-18
Add a `SlotUpkeep` type to enforce dependencies between rules

WhatisRT:
We want to ensure that the `Slot` rule only fires if all our block production duties are completed. To do this, I've added a set that keeps track of these to the state and adjusted the rules so that they can only fire one per slot. Then, `Slot` checks that all duties are indeed completed. This means that if we can't produce anything, we still need to update the state to not get stuck, which is why there are now negative versions of some rules.

While doing this, I noticed that we're not keeping the state of the base chain in our state and instead always universally quantifying over all those states. I fixed that too.

### DIFF
--- a/formal-spec/Leios/SimpleSpec.agda
+++ b/formal-spec/Leios/SimpleSpec.agda
@@ -22,6 +22,7 @@ module Leios.SimpleSpec (a : LeiosAbstract) (let open LeiosAbstract a) (let open
   (sk-IB sk-EB sk-V : PrivKey)
   (pk-IB pk-EB pk-V : PubKey)
   (let open Leios.Base a vrf') (B' : BaseAbstract) (BF : BaseAbstract.Functionality B')
+  (initBaseState : BaseAbstract.Functionality.State BF)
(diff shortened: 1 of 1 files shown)


formal-spec/Leios/SimpleSpec.agda
```diff
@@ -90,8 +99,14 @@ record LeiosState : Type where
   constructLedger : List EndorserBlock → List Tx
   constructLedger = concatMap lookupTxs
 
-initLeiosState : VTy → StakeDistr → LeiosState
-initLeiosState V SD = record
+  needsUpkeep : SlotUpkeep → Set
+  needsUpkeep = _∈ Upkeep
```
yveshauser, line 103: That should be `_∉`, shouldn't it?

yveshauser (approved):
lgtm (just the small thing mentioned above)

formal-spec/Leios/SimpleSpec.agda
WhatisRT: Ah, yes, thank you!

PR #76, opened 2024-11-20
Formal specification: Short Leios

yveshauser:
- Split `SimpleSpec` to support other variants of _Leios_, in particular _Short Leios_
- Added a new type  ` _⇉⋆_`  allowing to build traces for the different variants


formal-spec/Leios/Spec.agda
```diff
@@ -0,0 +1,255 @@
...
+                 ks ks' : K.State
+                 msgs   : List (FFDAbstract.Header ffdAbstract ⊎ FFDAbstract.Body ffdAbstract)
+                 eb     : EndorserBlock
+                 ebs    : List EndorserBlock
+                 txs    : List Tx
+                 V      : VTy
+                 SD     : StakeDistr
+                 pks    : List PubKey
+
+module Protocol (_⇀_ : LeiosState → LeiosState → Type) (allUpkeep : ℙ SlotUpkeep) (open Leios.Voting a) (va : VotingAbstract LeiosState) (open VotingAbstract va)
```
WhatisRT, line 181: I'm not sure if I like `⇀`. In the Ledger we're using this notation for maps, so it looks a lot like that to me. But it does make sense with the other arrows we have, and I don't see any other obvious candidate while staring at the options in the Agda input method.

WhatisRT (approved):
Looks great!

PR #77, opened 2024-11-20
Preparations for an executable specification

WhatisRT:
This PR bumps the ref of the prelude to include the newly upstreamed `Computational` type class and reorganizes the module parameters of the simple spec into a `SpecStructure`, so that it's easier to make new modules interacting with it and extend it in the future. There's also some other minor rearrangement and cleanup.


yveshauser (approved):
:+1:

PR #94, opened 2024-12-04
Voting and certificates section of first tech report

bwbush:


docs/technical-report-1.md
```diff
@@ -41,10 +41,65 @@
...
+- $l_\text{sec} = 80$
+- $n_\text{votes} = 500$
+- $2^{-l_\text{sec}} = 8.27 \cdot 10^{-25}$
+- $u_\text{ALBA} = 148$
+- probability of adversarial quorum
+  - 35% adversarial stake: $p = 1.71 \cdot 10^{-21}$
+  - 40% adversarial stake: $p = 7.69 \cdot 10^{-13}$
+  - 45% adversarial stake: $p = 2.87 \cdot 10^{-7}$
+- probability of honest quorum
+  - 35% adversarial stake: $p = 0.917$
```
amesgen, line 91: Sadly, I think that this probability is much smaller because there is another step with ALBA, namely to actually create the certificate/proof. The completeness property of ALBA (ie the probability that one is not able to create a proof is at most $2^{-\lambda}$) only applies to provers with at least $n_p = 0.92 \cdot n_\text{votes} = 460$ votes.

However, when we have $100\\% - 35\\% = 65\\%$ honest stake, then only $0.65 \cdot n_\text{votes} = 325$ votes are cast on average if the adversary simply abstains. However, this is significantly smaller than $n_p$, so one can't rely on ALBA completeness. One still can try to "optimistically" create a proof even when one has less than $n_p$, but more than $n_f$ elements (and indeed, the probability that one can still create a proof with very high probability if one has only slightly less than $n_p$ elements), but this probability decreases exponentially. See https://github.com/cardano-scaling/alba/discussions/17 for much more detail.

For the concrete ALBA parameters chosen here, the probability to create a valid proof with $0.65 \cdot n_\text{votes} = 325$ is at most `1.86e-20`, so it is extremely unlikely that one will ever be able to create an ALBA proof with a 35% adversary. I adapted the plot from https://github.com/cardano-scaling/alba/discussions/17 for the ALBA parameters proposed here:
optimistic-proving

In particular, even when the adversary only has 20% or 15% stake, so there are $0.8 \cdot n_\text{votes} = 400$ or $0.85 \cdot n_\text{votes} = 425$ votes if the adversary abstains, then the probability that one can create an ALBA proof out of these is still at most `4.13e-7` or `3.26e-3`, so still quite small (in particular much smaller than $p = 0.917$.

---

So one either needs to change ALBA parameters (e.g. making $n_p$ smaller, but then ALBA proofs get much larger), or fall back to some different scheme when there are few votes (in which case proofs also might get larger, although just when there is an adversary with some non-tiny amount of stake, or when honest participation is low for whatever reason).

(For Peras, all of these things are even more pronounced as $n_p$ needs to be set to at least 75% of the votes.)

---

<details>
<summary>matplotlib script for the graph above</summary>

```python
#!/usr/bin/env nix-shell
#! nix-shell -i python3 --pure
#! nix-shell -p "python3.withPackages (ps: [ ps.matplotlib ps.scipy ])"
#! nix-shell -p texlive.combined.scheme-full
#! nix-shell -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/4633a7c72337ea8fd23a4f2ba3972865e3ec685d.tar.gz

from math import *
import matplotlib.pyplot as plt
import numpy as np

plt.rcParams["text.usetex"] = True
plt.rcParams["text.latex.preamble"] = "\\usepackage{siunitx}"

lam = 80  # security parameter
n = 500
n_p_frac = 0.92
n_f_frac = 0.60
n_p = int(n_p_frac * n)
n_f = int(n_f_frac * n)

print(f"n_p = {n_p}, n_f = {n_f}, λ = {lam}")

assert n_f < n_p

# From Corollary 3:
u = ceil((lam + log2(lam) + 5 - log2(log2(e))) / log2(n_p / n_f))

# Make sure we are in Case 1 in the proof of Corollary 3.
S_1_bot = n_p / (17**2 / (9 * log2(e)) * u**2) - 7 < 1
S_2_bot = n_p / (17**2 / (9 * log2(e)) * u**2) - 2 < 1
assert S_1_bot or S_2_bot

# With the parameters above, we are in Case 1 in the proof of Corollary 3 (or
# the "≤ λ²" column in Figure 1).
d = ceil(32 * log(12) * u)
q = 2 * log(12) / d
r = ceil(lam)

print(f"u = {u}, d = {d}, q = {q}, r = {r}")

optimistic_prover_success_bound = lambda v: r * (v / n_p) ** u * d * q

fig, ax = plt.subplots(figsize=(9, 6))
X = np.arange(n_f - 10, n_p + 1)
ax.set_title(
    "Optimistic ALBA proving\n"
    f"$n_p = {n_p} = {n_p_frac} \\cdot n$ and $n_f = {n_f} = {n_f_frac} \\cdot n$ where $n = {n}$\n"
    f"$\\lambda = {lam}$\n"
)
ax.set_xlabel("number of elements $|S_p|$ the prover has access to")
ax.grid(True)

ax.set_yscale("log", base=10)
ax.plot(X, np.vectorize(optimistic_prover_success_bound)(X))
ax.set_ylabel("probability (upper bound) to create a valid proof")
ax.tick_params(axis="y")

secax = ax.secondary_yaxis(location="right")
secax.set_yscale("log", base=2)

ax.axvline(x=n_p, linestyle="--", color="red", label="$n_p$")
ax.axvline(x=n_f, linestyle="--", color="green", label="$n_f$")
ax.legend()
ax.set_ylim(top=1)

axratio = ax.secondary_xaxis(location="top", functions=(lambda v: v / n, lambda r: r * n))
axratio.set_xlabel("$|S_p| / n$")

for specific_x in [0.65 * n, 0.8 * n, 0.85 * n]:
    specific_y = optimistic_prover_success_bound(specific_x)
    ax.annotate(
        f"$\\num[exponent-mode = scientific, round-mode = places]{{{specific_y}}}$",
        xy=(specific_x, specific_y),
        xytext=(20, -3),
        textcoords="offset points",
        arrowprops=dict(facecolor="black"),
    )

fig.savefig("optimistic-proving.png", dpi=300, bbox_inches="tight")
```

</details>

bwbush: Thanks so much, this is very helpful. Aside from choosing different parameters, I see that I need to separate the general probabilistic computations about quorums (which would hold for any certificate scheme) from the additional considerations that a specific scheme like ALBA imposes. It feels like those general computations are an approximate _upper bound_ on the probabilities of a specific scheme.

We should probably clearly separate these cases, which aren't disjoint and which share similar computations.

| Content of certificate | Quorum on honest votes? | Certificate created? | Description | Dangerous? |
|---|---|---|---|---|
| honest votes | yes | yes | proper honest certificate | no |
| honest votes | no | yes | erroneous honest certificate | yes |
| honest votes | yes | no | failure to make honest certificate | yes |
| honest votes | no | no | no honest quorum | no |
| adversarial votes | yes | yes | adversarial certificate when honest one was possible | yes |
| adversarial votes | no | yes | adversarial certificate | yes |
| adversarial votes | - | no | failure to make adversarial certificate | no |

bwbush: @amesgen, in 0e94f71559e6e9a934017639b76de0b1e0e80f96 I revised the analysis for ALBA based on your helpful guidance and python code.

PR #97, opened 2024-12-11
Short-pipeline Leios updates

yveshauser:
The PR introduces a new type for data to be sent to the base chain. Rather that having either an EndorserBlock or a list of transactions, we allow as well both together. This new type for now is called `RankingBlock`


WhatisRT (approved):
LGTM

PR #99, opened 2024-12-12
docs/d2

will-break-it:
Add d2 diagrams for rs simulation from initial code review. It's a proposal and not final - hope it saves time for the review session or finds reuse for the demo.


.github/workflows/ci.yaml
```diff
@@ -51,7 +50,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04]
-        ghc-version: ['9.8']
+        ghc-version: ["9.8"]
```
bwbush, line 53: Our [CONTRIBUTING.md](https://github.com/input-output-hk/ouroboros-leios/blob/6e7dcaf37c6de16365e0e83deb71ad26e0fce781/CONTRIBUTING.md?plain=1#L71) is out of sync: it recommends GHC 9.6.3. I think both versions probably work, but we might want to align the two files.

bwbush (approved):
I found the diagram very helpful. When we're near release, we might want hyperlinks from the diagram boxes to the rust documentation or source code. For now, I think that's premature.

Thanks for adding the rust to the CI.

PR #117, opened 2024-12-19
Bounded parallelism and CPU Cores usage chart

Saizan:
* Added rather crude chart to show number of cores used per node.
* TODO: moving average might be more illustrative.
* Added `-N` flag to control parallelism, accepts 'unbounded' or number of cores.
* `cabal run ols -- viz short-leios-p2p-1 -N unbounded` shows a node using 13 cores around 30s.
* Otherwise things seem to get busy around votes, though maybe less so as time goes on?

### DIFF
--- a/simulation/ouroboros-leios-sim.cabal
+++ b/simulation/ouroboros-leios-sim.cabal
@@ -98,6 +98,7 @@ library
     VizSimRelayP2P
     VizSimTCP
     VizUtils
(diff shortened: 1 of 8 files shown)


wenkokke:
> Added -N flag to control parallelism, -1 unbounded, otherwise that many cores.

Why not use `0` or a `"unbounded"` for unbounded usage?

PR #128, opened 2025-01-10
Various fixes and other changes

WhatisRT:
This is everything in #111 except the actual reference implementation. Lots of small fixes here and there and the changes to voting we agreed on. I'll rebase #111 on top of this.


yveshauser (approved):
:+1:

PR #139, opened 2025-01-17
ci: only run ci for relevant changes

will-break-it:
- [x] `formal-spec-typecheck` should only be run if changes occur in the `formal-spec` directory
- [x] `simulation-test`, `simulation-hlint` and `simulation-fourmolu` only run if changes occurred in the `simulation` directory or in the data directory as it contains simulation inputs
- [x] `sim-rs-check` only runs if changes occurred in the top-level `sim-rs` or data directory
- [x] `docs-generate-d2-diagrams` only runs for changes in `d2` files
- [x] `docs-build` only runs if changes were made to the top-level `site` directory
- [x] `docs-publish` only runs if merge to main branch happens and changes in the `site` directory happened.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,6 @@ jobs:
   ################################################################################
   # Formal Specification in Agda - under /formal-spec/
   ################################################################################
-
   formal-spec-typecheck:
     name: "formal-spec: Typecheck"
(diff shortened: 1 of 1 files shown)


wenkokke:
@will-break-it would you be willing to add similar constraints on the other projects, assuming each only depends on its respective subdirectory?

PR #148, opened 2025-01-20
fix: move cpu core count into node instead of link

will-break-it:


wenkokke (changes requested):
See [my comment](https://github.com/input-output-hk/ouroboros-leios/pull/148#issuecomment-2603113525).

wenkokke:
Fixed this in 29f75e3a9429dd405616644b23b7ca0a73d09ad4 (which also adds formatting).

Apologies, I didn't see your update to the PR.

PR #153, opened 2025-01-21
Fix build with ghc-9.6.x

dcoutts:
* Some trickery with named fields. Fixed by an explict type signature.
* Unused type extensions.
* New library type aliases

In truth it doesn't build-as is because the custom version of io-sim doesn't build with ghc-9.6, but the custom version is not needed except for performance. Would be nice if we could get back to upstream io-sim.

### DIFF
--- a/simulation/src/ChanDriver.hs
+++ b/simulation/src/ChanDriver.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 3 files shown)


wenkokke:
Please hold off on merging this, since CI isn't running for Haskell changes at the moment (and it'll conflict with #151).

Saizan:
@dcoutts RE using an upstream io-sim, most of it is already upstream, but not in the release/1.5.0.0 maintenance branch and ouroboros-network depends on 1.5.*. A backport PR should not take much.
We would also want to add a ThreadId constructor that just takes an Int, hopefully that's easy to backport too.

Or, we get someone to bump ouroboros-network to io-sim-1.6?

PR #173, opened 2025-01-27
simulation: make hlint report only about restricted modules

Saizan:
I realized hlint was causing me unneeded friction, so figured I would tailor the config to its intended purpose.


wenkokke:
Have you tested that this works?

Saizan:
Yes, hlint does complain if you add an import it should warn about

PR #179, opened 2025-02-04
delta-q: handle redundant x values in compact

SupernaviX:
If we create a CDF with multiple points at the same X value, only the highest Y for each X is relevant.


rkuhn (approved):
While these changes are good, they won’t fix the current issue: `compact` isn’t called when constructing a CDF. But it probably should, so I’ll add that. The library is currently in a state following from some nearly finished experimentation, it would need some API cleanup to make it actually usable as a library.

SupernaviX:
`compact isn’t called when constructing a CDF`

It isn't called during direct construction, but it is in [the code path which the topology checker uses](https://github.com/input-output-hk/ouroboros-leios/blob/57ed92deac9ee0626fdf8e8bb247dcaf847cfed0/delta_q/src/cdf.rs#L282-L288). So this change does fix my immediate issue.

PR #193, opened 2025-02-14
Update sizes and timings in default config

Saizan:
Most size and timing parameters in the shared configuration are expected to not vary between different the scenarios we will explore, so it makes sense to agree on sensible default values.

The proposed changes are based on info from `crypto-benchmarks.rs`, the Short Leios spec, and `cardano-node`, the inline comments have more detailed references and rationales.


data/simulation/config.default.yaml
```diff
@@ -110,9 +170,17 @@ vote-diffusion-max-window-size: 100
...
-cert-generation-cpu-time-ms-constant: 50.0
-cert-generation-cpu-time-ms-per-node: 1.0
-cert-validation-cpu-time-ms-constant: 50.0
-cert-validation-cpu-time-ms-per-node: 1.0
-cert-size-bytes-constant: 32
-cert-size-bytes-per-node: 32
+# vote-spec#"certificate bytes"
+cert-size-bytes-constant: 136
+# vote-spec#"certificate bytes" ((80/8) + 76 * (100 - 80))/100
+cert-size-bytes-per-node: 15
```
bwbush, line 176: I think this is fine for the time being.

bwbush (approved):
Looks great. This will give us reasonably accurate accounting.

data/simulation/config.default.yaml
```diff
@@ -42,29 +42,54 @@ tx-max-size-bytes: 16384
...
+# VRF proof   80
+# Body hash   32
+# RB Ref      32
+# Signature   64
+# Total       304
+#
+# NOTE: using a KES Signature (like for Praos headers)
+#       would instead more than double the total to 668.
+#       And even 828 including Op Cert.
+ib-head-size-bytes: 304
```
Saizan, line 82: @pagio do you believe we want KES signatures in IB headers and EBs, or something smaller would be enough?

PR #194, opened 2025-02-14
data: add trace schema for haskell & rust

will-break-it:
- Draft of current simulation trace schema
- Serves as base for discussed changes


Saizan:
@WhatisRT @yveshauser could you comment on whether the haskell trace schema contains enough information for conformity testing, and what else is needed otherwise?

SupernaviX:
@will-break-it I have PR #206 ready to merge either before or after this one, to make the rust sim actually match this spec

PR #202, opened 2025-02-18
Trace Schema Changes

will-break-it:
- [x] add cpu_time_s field to track total CPU time across cores
- [x] align block event names
- [x] separate block lifecycle actions from network actions

### DIFF
--- a/data/simulation/trace.haskell.d.ts
+++ b/data/simulation/trace.haskell.d.ts
@@ -41,6 +41,8 @@ interface EndorserBlockEvent extends BaseBlockEvent {
 
 interface RankingBlockEvent extends BaseBlockEvent {
     slot: number;
+    endorse_blocks: string[]; // References to certified endorser blocks
+    payload_bytes: number; // Size of directly included transactions
 }
 
 interface VoteEvent extends BaseBlockEvent {
--- a/data/simulation/trace.haskell.d.ts
+++ b/data/simulation/trace.haskell.d.ts
@@ -13,7 +13,7 @@ type HaskellEvent =
     | HaskellNetworkEvent; // Combine Sent/Received into network events
 
 interface HaskellCpuEvent {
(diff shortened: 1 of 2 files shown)


SupernaviX (approved):

data/simulation/trace.haskell.d.ts
```diff
@@ -4,85 +4,67 @@ export interface HaskellTraceEvent {
...
+    slot: number;
+    input_blocks: string[]; // References to input blocks
 }
 
-interface HaskellGeneratedRBEvent extends HaskellGeneratedBaseEvent {
+interface RankingBlockEvent extends BaseBlockEvent {
     kind: "RB";
-    id: string;
+    slot: number;
 }
```
Saizan, line 47: I believe we want to include two more fields:
-  `endorse_blocks : string[]`  -- which certified EBs are included (at most 1 in ShortLeios though)
- `tx_payload_bytes : number` -- whether any txs are directly included in the block body.

data/simulation/trace.haskell.d.ts
```diff
@@ -4,85 +4,67 @@ export interface HaskellTraceEvent {
...
 
-interface HaskellGeneratedVTEvent extends HaskellGeneratedBaseEvent {
+interface VoteEvent extends BaseBlockEvent {
     kind: "VT";
-    id: string;
     votes: number;
-    endorse_blocks: string[];
+    endorse_blocks: string[]; // References to endorser blocks
```
Saizan, line 52: I think we want `slot: number` here too, maybe it can go into BaseBlockEvent? Btw `slot` will be the one the slot is generated for.

Saizan:
@will-break-it is it intentional that the data for the block (e.g. `payload_bytes` for IBs) is repeated in both *Generated and *Received events? Not a problem, just confirming.

PR #218, opened 2025-02-27
fix: optimize ib size for 3 TCP round trips

will-break-it:
During TCP's slow start phase, the congestion window (CWND) typically:

Starts with an initial congestion window (initcwnd) of 10 MSS (Maximum Segment Size)
Doubles after each round trip time (RTT)

With a standard MSS of 1460 bytes:

- Round Trip 1: 10 MSS = 10 × 1460 = 14,600 bytes
- Round Trip 2: 20 MSS = 20 × 1460 = 29,200 bytes
- Round Trip 3: 40 MSS = 40 × 1460 = 58,400 bytes

Total for 3 round trips: 14,600 + 29,200 + 58,400 = 102,200 bytes

However, we need to account for TCP/IP overhead:

- TCP header: 20 bytes
- IP header: 20 bytes

This means each segment carries 1460 bytes of payload but requires 1500 bytes on the wire.
When accounting for this overhead and some implementation variations (like initial window sizes that might be slightly different in various TCP stacks), the practical optimal size for 3 round trips is approximately 98304 bytes (96KB).

This value represents a sweet spot that:
1. Fully utilizes 3 complete round trips
2. Doesn't require a partial 4th round trip
3. Accounts for real-world TCP implementation variations

### DIFF
--- a/data/simulation/config.default.yaml
+++ b/data/simulation/config.default.yaml
@@ -81,8 +81,8 @@ ib-shards: 1
 #       would instead more than double the total to 668.
 #       And even 828 including Op Cert.
 ib-head-size-bytes: 304
-# 100kB, using praos max size as ballpark estimate.
-ib-body-avg-size-bytes: 102400
+# 98KB to optimize for 3 TCP round trips
+ib-body-avg-size-bytes: 98304
 ib-body-max-size-bytes: 327680
 # Here we also use praos blocks as ballpark estimate.
 # Sec 2.3 Forging, of the benchmark cluster report, lists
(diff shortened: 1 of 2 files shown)


Saizan:
FWIW, that was the original default ib-body-size default (taken from the praos simulation body default). It got lost when moving to the shared configuration. When revamping the defaults I went for a rounder number but now the specific 96kB makes more sense.

will-break-it:
@Saizan so I'll merge it then?

Saizan:
@will-break-it probably warranted and limited harm I would say. I'm no authority on tcp matters though.

bwbush (approved):

PR #225, opened 2025-02-28
Updating log format to new schema

Saizan:
@will-break-it how much are we gaining from the new schema if there isn't an actual shared subset between Haskell and Rust?


will-break-it:
My initial goal was to read it in to derive grammar rules for the logs. But at this point I think it's secondary - it would be great to have the schema reflect the current format of the each sim's log though. But they don't have to completely align.

If we run into issues/ difficulties comparing the sims due to the format, we can revisit this.
Does that sounds fair?

PR #244, opened 2025-03-12
Update documentation for: FAQ, Resources, and How It Works pages

kbennett2000:
Made documentation updates to the FAQ, Resources, and How It Works pages.


will-break-it:
Can you remove the `package-lock.json` files - they shouldn't be committed. Perhaps we can add them to the `.gitignore`

will-break-it (changes requested):
Aside from the unneeded lock files, looks good!

Quantumplation:
Ah right lol

PR #263, opened 2025-03-20
Mapping of simulation log to Agda data types

yveshauser:
This PR introduces:
- Trace verification (from [ouroboros-leios-formal-spec](https://github.com/input-output-hk/ouroboros-leios-formal-spec))
- Use of `leios-trace-hs` Haskell module for reading log and configuration files


leios-trace-verifier/Parser.agda
```diff
@@ -70,35 +83,118 @@ record TraceEvent : Type where
...
-open import Leios.Trace.Verifier
 
-open import Leios.Defaults 2 fzero using (st)
+open import Leios.Defaults 10 fzero hiding (LeiosInput; LeiosOutput; SLOT)
+open import Leios.Trace.Verifier 10 fzero
 open import Leios.Short st
+import Data.String as S
+
+nodeId : String → Fin 10
+nodeId "node-0" = fzero
```
WhatisRT, line 93: I think you can write `# 0`, etc.

yveshauser: As the number of nodes (here 10) is now read from the configuration file, this function has become more generic as well

PR #272, opened 2025-03-28
simulation: docs: regenerated topologies and .csv

Saizan:


Saizan:
@wenkokke I was curious how things would look so I used the new default topology to rerun the scenarios for the realism report. Not a lot seems to change, but I only briefly skimmed the plots.

PR #299, opened 2025-04-16
Epoch-utilization, fee, and reward computations

bwbush:
The analysis [Motivating Leios: Fees and rewards](https://github.com/input-output-hk/ouroboros-leios/blob/bwbush/epoch/analysis/epoch-utilization.md) estimates the transaction throughput needed to maintain current reward levels on `mainnet` in the future. It assumes that fee-related protocol parameters remain the same in the future.


will-break-it (approved):
👍

PR #300, opened 2025-04-17
Costs

will-break-it:
- [x] refactored previous calculations/ estimates for storage, CPU and network egress.
- [x] added IOPS & memory usage
- [x] added aggregated cost table
- [x] added aggregated cost/revenue table


docs/cost-estimate/01-compute-cpu.md
```diff
@@ -0,0 +1,236 @@
...
+| Vote        | ~150 bytes       | 0.816ms              | `vote-validation-cpu-time-ms`          |
+| Certificate | 7,168 bytes      | 140ms                | `cert-validation-cpu-time-ms-constant` |
+
+### Base Parameters
+
+| Parameter       | Value          | Source                        |
+| --------------- | -------------- | ----------------------------- |
+| Stage length    | 20 slots (20s) | `leios-stage-length-slots`    |
+| EBs per stage   | 1.5            | `eb-generation-probability`   |
+| RBs per stage   | 1              | Protocol parameter            |
```
bwbush, line 73: This might be confusing because it makes it sound like Praos blocks are controlled by Leios stages.

docs/cost-estimate/02-compute-ram.md
```diff
@@ -0,0 +1,185 @@
...
+| Component              | Size Range   | Scaling Behavior                  |
+| ---------------------- | ------------ | --------------------------------- |
+| Base Node Process      | 500 MB       | Fixed                             |
+| Hot UTxO Cache         | 200-500 MB   | Configurable, minimal with IB/s   |
+| Indexes & Filters      | 100-300 MB   | Logarithmic with ledger size      |
+| Write Buffers          | 100-500 MB   | Linear with transaction rate      |
+| Stake Data             | 200-500 MB   | Linear with stake pool count      |
+| Vote Processing        | 50-100 MB    | Minimal with IB/s                 |
+| Certificate Processing | 20-40 MB     | Minimal with IB/s                 |
+| Consensus Overhead     | 100-300 MB   | Varies with network parameters    |
```
bwbush, line 85: Nice!

docs/cost-estimate/03-storage.md
```diff
@@ -0,0 +1,224 @@
...
+> **100% filled blocks assumption:**
+>
+> This analysis assumes fully utilized block space for both Ouroboros Praos and
+> Leios protocols, similar to the egress analysis. We'll calculate storage
+> requirements at different IB/s rates.
+>
+> **Compression note:** While early experiments show that >70% compression is
+> feasible on mainnet data, this analysis does not include compression benefits.
+> The calculations represent raw storage requirements without compression.
+
```
bwbush, line 14: I wonder if we should note that these costs increase each month as the ledger gets larger. In some sense the units are "(USD/ month) / month" because this is the *velocity* of the cost increase.

docs/cost-estimate/05-iops.md
```diff
@@ -0,0 +1,203 @@
...
+and ledger state updates.
+
+### Block Storage Operations
+
+| Operation Type | IOPS per Block | Notes |
+| -------------- | -------------- | ----- |
+| Block write    | 1              | Writing a validated block to disk |
+| Block read     | 0.1            | Reading blocks for chain verification (10% read rate) |
+| State update   | 3              | Ledger state updates per block |
+| **Total**      | **4.1**        | IOPS per block |
```
bwbush, line 25: My understanding is that IOPS are computed as writes/reads of ***filesystem blocks***, not logical operations like writing/reading a Praos/Leios block. AWS EBS seems to have a 4 kiB block size for the volumes that a cardano node would use.

So, to compute IOPS costs, we need to convert from Praos/Leios reads/writes to filesystem blocks.

docs/cost-estimate/05-iops.md
```diff
@@ -0,0 +1,203 @@
...
+
+> [!Note]
+> IOPS per Block values are derived from empirical measurements of the Cardano node's 
+> disk activity during block validation and chain synchronization. The state update 
+> value (3 IOPS) represents the average database operations needed to apply a typical 
+> block to the ledger state.
```
bwbush, line 31: My understanding is that IOPS are computed as writes/reads of ***filesystem blocks***, not logical operations like writing/reading a Praos/Leios ~~block~~. AWS EBS seems to have a 4 kiB block size for the volumes that a cardano node would use.

So, to compute IOPS costs, we need to convert from Praos/Leios reads/writes to filesystem blocks.

docs/cost-estimate/05-iops.md
```diff
@@ -0,0 +1,203 @@
...
+|------|---------|-------|---------|-------|---------|-------|--------------|
+| 0.05 | 0.205   | 50.6% | 0.09    | 22.2% | 0.11    | 27.2% | 0.405        |
+| 1    | 4.1     | 95.3% | 0.09    | 2.1%  | 0.11    | 2.6%  | 4.3          |
+| 5    | 20.5    | 99.0% | 0.09    | 0.4%  | 0.11    | 0.5%  | 20.7         |
+| 10   | 41      | 99.5% | 0.09    | 0.2%  | 0.11    | 0.3%  | 41.2         |
+| 20   | 82      | 99.8% | 0.09    | 0.1%  | 0.11    | 0.1%  | 82.2         |
+| 30   | 123     | 99.8% | 0.09    | 0.1%  | 0.11    | 0.1%  | 123.2        |
+
+This shows that IB processing dominates the IOPS requirements at rates above 1 IB/s, with
+EB and RB operations contributing only marginally to the overall IOPS requirements.
```
bwbush, line 169: Is this consistent with the measurements below on a `mainnet` node running in steady state (i.e., not syncing)?

image

docs/cost-estimate/README.md
```diff
@@ -0,0 +1,200 @@
...
+|                  | IOPS                | $7.00       | $7.00       | $10.00        | $20.00        | $40.00        | $60.00        |
+|                  | **Total (Linode)**  | **$64.75**  | **$92.41**  | **$161.69**   | **$365.72**   | **$746.61**   | **$1,067.50** |
+| **Hetzner**      | Compute (vCPU)      | $5.40       | $5.40       | $17.80        | $17.80        | $32.90        | $32.90        |
+|                  | Compute (RAM)       | $4.98       | $9.96       | $9.96         | $19.93        | $19.93        | $19.93        |
+|                  | Storage             | $0.60       | $11.16      | $55.58        | $111.11       | $222.17       | $333.24       |
+|                  | Egress              | $0.00       | $0.00       | $6.43         | $12.87        | $25.70        | $38.54        |
+|                  | IOPS                | $5.00       | $5.00       | $8.00         | $15.00        | $30.00        | $45.00        |
+|                  | **Total (Hetzner)** | **$15.98**  | **$31.52**  | **$97.77**    | **$176.71**   | **$330.70**   | **$469.61**   |
```
bwbush, line 64: These results are consistent with the online Leios calculator if one approximately aligns the assumptions.

docs/cost-estimate/README.md
```diff
@@ -0,0 +1,200 @@
...
+The following table maps IB/s rates to transactions per second (TPS) for each transaction type:
+
+| IB/s Rate | Average Tx (1,400 bytes) | Minimum Tx (250 bytes) | Maximum Tx (16,000 bytes) |
+|-----------|--------------------------|------------------------|---------------------------|
+| 0.05      | 3.5 TPS                  | 19.65 TPS              | 0.3 TPS                   |
+| 1         | 70 TPS                   | 393 TPS                | 6 TPS                     |
+| 5         | 350 TPS                  | 1,965 TPS              | 30 TPS                    |
+| 10        | 700 TPS                  | 3,930 TPS              | 60 TPS                    |
+| 20        | 1,400 TPS                | 7,860 TPS              | 120 TPS                   |
+| 30        | 2,100 TPS                | 11,790 TPS             | 180 TPS                   |
```
bwbush, line 126: Interesting! and consistent with other estimates that we've made previously.

docs/cost-estimate/03-storage.md
will-break-it: I have such a note in the README but can add it here as well or would you rather change the unit/ or make a annual forecast for the next couple years?

> [!Important] 
> Storage costs shown above represent only one month of
> blockchain data. As blockchain history accumulates over time, storage
> requirements and associated costs will continuously increase, particularly for
> higher IB/s rates.

https://github.com/input-output-hk/ouroboros-leios/pull/300/files#diff-098f39927c86fcfe24b0bdbf943cee7bfb7385c8ff3496e1cd754ef66a93d758R72-R76

docs/cost-estimate/05-iops.md
will-break-it: You're right, I used the read and write numbers from the UTxO HD paper and didn't relate it to filesystem blocks. I'll have to refactor this.

docs/cost-estimate/01-compute-cpu.md
will-break-it: I've removed it.

docs/cost-estimate/05-iops.md
will-break-it: Mh even my refactored numbers for Praos don't match this if I read it correctly. However, I focused only on the core blockchain operations (block writes/reads and state updates) related to consensus, while a running node performs many additional IO operations:

- Database Operations: Continuous UTxO set access, indexing, and maintenance
- Mempool Management: Transaction validation and storage
- Chain Selection: Reading multiple blocks for fork choice
- Network Operations: Temporary storage of data during propagation
- Background Maintenance: Database optimization, garbage collection
- Logging: Extensive logging of node activities
- Operating System and Runtime: Swap, OS-level caching, etc.

The real-world IOPS are approximately 85-225× higher than what I currently have as theoretical minimum calculations.

docs/cost-estimate/03-storage.md
bwbush: I think it's worthwhile to note it in both places.

docs/cost-estimate/05-iops.md
bwbush: These are really hard to estimate because it is so implementation dependent. It's not a major component of the cost, so I don't think we need to spend time on making this too accurate.

PR #386, opened 2025-06-05
Differentiate between duplicate, invalid and conflicting txs

ch1bo:
We just discussed that the Cardano p2p network / tx diffusion protocol would cover for duplicate or invalid transactions and only the conflicting txs (e.g. multiple valid txs spending the same UTxO) may be an issue.


will-break-it:
I think, what's most interesting to me is to explore **alternative** mitigation possibilities with different trade-offs.
And we currently have some - like the tomb-stoning extension for duplicates.

bwbush (approved):

PR #401, opened 2025-06-11
CIP section on metrics

bwbush:


docs/cip/README.md
```diff
@@ -247,6 +247,91 @@ As a result of this decoupled approach, Leios can utilize nearly the full bandwi
...
+
+$$`
+\epsilon_\text{temporal}(s) = \text{fraction of transactions included in the ledger within } s \text{ slots of their inclusion in a local memory pool}
+`$$
+
+***Network efficiency:*** Effective utilization of the network can be characterized by the ratio of bytes of transactions reaching the ledger to the average network traffic per node.
+
+$$`
+\epsilon_\text{network} = \frac{(\text{bytes of valid transactions reaching the ledger}) \cdot (\text{number of nodes in the network})}{\text{total bytes of network traffic}}
+`$$
```
ch1bo, line 287: Would it make sense to make this a "local metric"? i.e. bytes of txs in the ledger divided by bytes processed by the protocol

That would make it somewhat easier to estimate and even measure locally?

docs/cip/README.md
```diff
@@ -247,6 +247,91 @@ As a result of this decoupled approach, Leios can utilize nearly the full bandwi
...
+
+***TX collision:*** Because Leios uses local sortition for electing which nodes produce IBs, there is a finite chance that a transaction will be included in more than one IB during the concurrency period when nodes are not aware of which transactions other nodes are including in their new IBs. The transaction collision probability is simply the fraction of transactions that appear multiple times in IBs.
+
+$$`
+p_\text{collision} = \text{fraction of transactions that appear multiple times in IBs}
+`$$
```
ch1bo, line 293: > fraction of transactions that appear multiple times in IBs

Sounds like this is about duplicates? I would propose to call this $p_\text{duplicate}$. It is very related to spatial and network efficiency as duplicates can clearly not provide value other than not to exist.

docs/cip/README.md
```diff
@@ -247,6 +247,91 @@ As a result of this decoupled approach, Leios can utilize nearly the full bandwi
...
+
+$$`
+\tau_\text{inclusion} = \text{mean number of slots for a transaction to be included in any IB}
+`$$
+
+***Voting failure:*** An unlucky set of VRF evaluations might result in insufficient voters being selected in a given pipeline, thus making it impossible to certify an EB in that pipeline.
+
+$$`
+p_\text{noquorum} = \text{probability of sufficient voters to achieve a quorum in a given pipeline}
+`$$
```
ch1bo, line 305: Is this only about the VRF probability or do we care about general likelihood of not certifying an EB in a pipeline?

docs/cip/README.md
```diff
@@ -247,6 +247,91 @@ As a result of this decoupled approach, Leios can utilize nearly the full bandwi
...
+
+***Peak CPU usage:*** Because CPU usage varies depending upon the node's activity, the maximum number of virtual CPU cores utilized by the protocol during any slot, $`\hat{q}_\text{vcpu}`$, provides a useful indication of computational burstiness and of how a virtual machine should be sized for Leios.
+
+***Bandwidth:*** If the bandwidth for inter-node communication drops below a given value, then the throughput of Leios (at a given level of demand) will be drop, as network congesting occurs.
+
+$$`
+\eta_\text{bandwidth}(b) = \frac{\text{bytes of transactions reaching the ledger if links have bandwidth } b}{\text{bytes of transactions reaching the ledger if bandwidth were infinite}}
+`$$
+
+***Conflict:*** Theoretically, an adversary could poison the global memory pool by simultaneously submitting mutually conflicting transactions to every node in the network. This would result in massive conflicts in the contents of IBs produced during the concurrency window, resulting in much wasted space in IBs, effectively reducing the throughput of the protocol.
```
ch1bo, line 323: This is quite hypothetical and sounds more like an attack vector than a metric. I would suggest to establish a corresponding metric to the duplicate txs one above, only about conflicts

docs/cip/README.md
```diff
@@ -247,6 +247,91 @@ As a result of this decoupled approach, Leios can utilize nearly the full bandwi
...
+
+$$`
+\hat{\eta}_\text{conflict} = \frac{\text{bytes of valid transactions reaching the ledger}}{\text{bytes of transactions reaching the ledger if no conflicts were present}}
+`$$
+
+***Message loss:*** Using the same metric $`\eta`$, one can also measure the loss in throughput due to unreliable network infrastructure where the fraction $`\ell`$ of messages sent from a node are lost before reaching their destination node.
+
+$$`
+\eta_\text{message}(\ell) = \frac{\text{bytes of transactions reaching the ledger if } \ell \text{ fraction of messages are lost}}{\text{bytes of transactions reaching the ledger if no messages are lost}}
+`$$
```
ch1bo, line 333: Why should we need this? There is no message loss in our network protocols (as all of them are pull-based and a node would re-request incompletely downloaded data).

ch1bo (commented):
> Are these formulated precisely enough?

Yes. The level of detail seems good to me.

> Are some unnecessary, either from the engineering or stakeholder perspective?

See my comments above

> What is missing?

Maybe something about exposure to stake-based attackers in the resilience section? e.g. throughput loss due to 15% stake adversary not following the protocol

Maybe I do understand now that message loss metric where you propose resilience against a network attacker?

will-break-it (approved):

PR #413, opened 2025-06-17
Add leios-late-ib-inclusion to the Haskell simulator

### LINKED ISSUE #410: Add LateIbInclusion to Haskell simulator
In the past month or so, Giorgos and Simon added a feature to the Short-to-Full extensions and the Rust simulator in which an EB also includes IBs from some recent iterations in addition to IBs from the current iteration.

For reference, the feature is controlled by this config file option in the Rust simulator: https://github.com/input-output-hk/ouroboros-leios/blob/f20c11d0d154fdcfa1a0df862438eaa1c912a5a4/data/simulation/config.d.ts#L41-L46

The Short-to-Full extension specification Google Doc says to include IBs from the "last 3 iterations", but my interpretation (especially due to the relevant figure in that Doc) is that one of those three is the current iteration. (So this feature results in EBs including 3x as many IB IDs rather than 4x as many.)

This needs to be added to the Haskell simulator.

nfrisby:
Closes #410.


simulation/src/LeiosProtocol/Short.hs
```diff
@@ -111,6 +111,10 @@ data LeiosConfig = forall p. IsPipeline p => LeiosConfig
   , variant :: LeiosVariant
   , headerDiffusionTime :: NominalDiffTime
   -- ^ Δ_{hdr}.
+  , leiosLateIbInclusion :: Bool
+  -- ^ Whether an EB also includes IBs from the two previous iterations.
+  --
+  -- TODO Merely one previous iteration if 'pipeline' is 'SingleVote'?
```
nfrisby, line 117: The Haskell simulator is configurable to use a single stage for voting )Vote) instead of two stages (VoteSend and VoteRecv).

The Short-to-Full spec only considers the two stage variant, when it specifies to include 3 iterations' IBs in each EB.

Would the Short-to-Full spec only include 2 iterations' IBs if Vote were a single stage? Should the Haskell simulator do that? @pagio @SupernaviX

nfrisby:
FYI, I did some force pushes that.

- fixed a bug
- added tests (very similar to implementation, but that's largely because this first implementation is unoptimized)
- ran `fourmolu` locally

data/simulation/config.d.ts
```diff
@@ -39,10 +39,10 @@ export interface Config {
...
   "leios-vote-send-recv-stages": boolean;
   /**
-   * Extends Leios so that EB producers include IBs directly from previous pipelines
-   * where no certified EB was observed.
-   * 
-   * Only supported by Rust simulation. */
+   * Extends Leios so that EB producers include IBs directly from previous pipelines.
+   * Due to casuality, the EB must always include them, even if those IBs end up being
+   * certified in their own pipeline.
+   */
```
SupernaviX, line 45: Taking another look at the extension, the behavior you describe looks correct. It's also the behavior that the rust simulation follows. I don't know why I wrote this comment.

nfrisby: If it were IBs from more than 3 iterations, then we _could_ start making decisions based on whether we see the EB. However, it's not clear we'd want to do that even though we could.

nfrisby:
I'm now expecting CI to build green.

simulation/src/LeiosProtocol/Short.hs
```diff
@@ -680,6 +686,26 @@ inputBlocksToEndorse cfg@LeiosConfig{pipeline = _ :: SingPipeline p} current buf
         , receivedBy
         }
 
+-- | Invokes 'inputBlocksToEndorse1' as many times as 'lateIbInclusion'
```
nfrisby, line 689: TODO this isn't going to work, since the IBs will have been pruned.

I'll need to relax the pruning logic https://github.com/input-output-hk/ouroboros-leios/blob/b5dd5b4440b01d79960bf6de6573dcecfbce8d3a/simulation/src/LeiosProtocol/Short/Node.hs#L537

nfrisby:
I converted this to a Draft PR because my study of the whole simulator has turned up at least three reasons this current diff is insufficient (see https://github.com/input-output-hk/ouroboros-leios/pull/413#discussion_r2155064821).

simulation/test/Main.hs
```diff
@@ -9,4 +10,5 @@ main =
   defaultMain . testGroup "ouroboros-leios-sim" $
     [ Test.Topology.tests
     , Test.Config.tests
+    , Test.ShortToFull.tests
```
bwbush, line 13: Hooray for test cases! :raised_hands:

bwbush (commented):
I don't have a global enough understanding of the implementation's inclusion logic to fully assess this, but changes seem to have the basics correct.

I'd very much like it if we could cite the specs as comments in the Haskell and Rust code. That direct link would make reviews much easier. Our use of google docs is not aligned very well with this style of commenting/citation.

simulation/test/Test/ShortToFull.hs
```diff
@@ -0,0 +1,79 @@
...
+    | ibTable' <- shrinkList (const []) testSetup.ibTable
+    ]
+
+test_leiosLateIbInclusion :: Test_leiosLateIbInclusion -> Property
+test_leiosLateIbInclusion testSetup =
+  Q.counterexample (show (cfg.sliceLength, iterations))
+    $ Q.counterexample ("on " <> show (map (.num) on))
+    $ Q.counterexample
+      (unlines $ ("off" :) $ map (show . map (.num) . off) $ iterations)
+    $ on Q.=== concatMap off iterations
```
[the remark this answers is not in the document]
nfrisby: For what it's worth, after my latest work on this branch, the implementation is no longer so similar to this test's formulation.

simulation/src/LeiosProtocol/Short/Node.hs
```diff
@@ -964,9 +995,6 @@ mkBuffersView cfg st = BuffersView{..}
     let tryCertify eb = do
           Certified{cert} <- Map.lookup eb.id votesForEB
           guard (not $ Set.member (endorseBlockPipeline cfg.leios eb) pipelinesInChain)
-          -- Note: we expect to have received the IBs for any
```
nfrisby, line 967: @SupernaviX When issuing a Ranking Block, does the Rust simulator skip over certs if it doesn't already have all of the IBs that the cert transitively includes?

nfrisby: Resolving this Conversation since we're discussing this elsewhere (ie on PR 427).

nfrisby:
I rebased so that the `SimulatorModel.md` is in-place and also amended it according to the changes in this PR.

I'm Merging now. Beyond the code review and my ad-hoc manual tests, the idea is that the Agda trace-verifier and/or confusing disagreements with the Rust simulator will eventually any catch bugs in this PR.

Issue #387, opened 2025-06-06
`EBGenerated` event needs an additional field for EndorserBlocks for Full-Short variant


yveshauser:
The Haskell library `leios-trace-hs` in the module `LeiosEvents` defines the domain model of the events written to the trace log.

The event `EBGenerated` needs an additional field for EndorserBlocks to cover pipeline referencing as described in the Full-Short variant of Leios

PR #416, opened 2025-06-18
Add EB links to EBGenerated event

nfrisby:
Closes https://github.com/input-output-hk/ouroboros-leios/issues/387.

### DIFF
--- a/data/simulation/trace.shared.d.ts
+++ b/data/simulation/trace.shared.d.ts
@@ -71,6 +71,7 @@ interface GeneratedEndorserBlock extends GeneratedBlockEvent {
     id: string;
     pipeline: number;
     input_blocks: BlockRef[];
+    endorser_blocks: BlockRef[];
 }
 
 interface GeneratedVote extends GeneratedBlockEvent {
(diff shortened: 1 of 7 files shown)


nfrisby:
I set this PR to be a Draft PR because `EBGenerated` is an event in the shared schema. I'm not sure if that needs to be updated as part of this PR, and if so, I think @SupernaviX would need to push a commit adding it to the Rust simulator as well?

nfrisby:
@yveshauser do you have any advice about how to fix this CI failure on this PR? Perhaps it'd make sense for you to push an Agda commit to it? https://github.com/input-output-hk/ouroboros-leios/actions/runs/15734666849/job/44344067399?pr=416

nfrisby:
- I renamed it to `endorser_blocks`, since I just found the parsing code in the `trace-processor` code that anticipates a field of that name.
- I added my best-guess of the necessary change to `leios-trace-verifier:Spec.Transition`. But that code is not included in the `cabal.project` file, and I don't know why not.

nfrisby:
I updated the schema files and rebased onto `main`.

SupernaviX (approved):

PR #420, opened 2025-06-19
CDDL

will-break-it:
- [x] Ranking Blocks
- [x] Votes & Certificates
- [x] Endorser Blocks
- [ ] Input Blocks
  - [x] Common/ Base
  - [x] Sharded
  - [ ] Over-collateralized
- [ ] Transactions
  - [x] Sharded
  - [ ] Over-collateralized


docs/cddl/diffs/common/votes-certificates.md
```diff
@@ -0,0 +1,137 @@
...
+  , ? aggregate_elig_sig   : bls_signature                           ; Optional aggregate eligibility signature
+  , aggregate_vote_sig     : bls_signature                           ; Aggregate BLS signature on (election_id || endorser_block_hash)
+  ]
+```
+<sub>[1] [Certificate Reference Implementation](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/cert.rs#L13-L21), [2] [Certificate Abstract Interface](https://github.com/input-output-hk/ouroboros-leios-formal-spec/blob/main/formal-spec/Leios/Base.agda#L24-L28)</sub>
+
+## Vote Structure
+
+The Leios voting system supports two types of voters: persistent voters (selected per epoch) and non-persistent voters (selected per election via local sortition).
+
```
bwbush, line 25: Should we note that votes would never appear on the ledger or in persistent disk?

docs/cddl/diffs/common/votes-certificates.md
```diff
@@ -0,0 +1,137 @@
...
+; Vote cast in a Leios election
+leios_vote = persistent_vote / non_persistent_vote
+
+; Vote cast by a persistent voter (90 bytes)
+persistent_vote =
+  [ 0                        ; Vote type identifier for persistent voter
+  , election_id              ; 8-byte election identifier  
+  , endorser_block_hash      ; 32-byte hash of the endorser block being voted on
+  , persistent_voter_id      ; 2-byte epoch-specific pool identifier
+  , vote_signature           ; 48-byte BLS signature on (election_id || endorser_block_hash)
```
bwbush, line 36: The Rust and Haskell simulations use a different data structure called a "vote bundle" because they vote on many EBs in the pipeline. Instead of single `endorser_block_hash` and `vote_signature` entries, they have a `map` from `endorser_block_hash` to `vote_signature`.  This saves ~10 bytes per persistent vote. Do we want to adopt this in the spec, or not worry about the 10 bytes?

https://github.com/input-output-hk/ouroboros-leios/blob/b5dd5b4440b01d79960bf6de6573dcecfbce8d3a/sim-rs/sim-core/src/model.rs#L208-L212

docs/cddl/diffs/common/votes-certificates.md
```diff
@@ -0,0 +1,137 @@
...
+; Total size: 28 + 96 + 96 + 448 = 668 bytes
+```
+<sub>[1] [Registration Struct](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/key.rs#L156-L162), [2] [Proof Generation](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/bls_vote.rs#L19-L23)</sub>
+
+## Cryptographic Types
+
+```cddl
+; Core BLS cryptographic primitives
+bls_signature           = bytes .size 48                             ; BLS12-381 G1 signature (compressed)
+bls_vkey                = bytes .size 96                             ; BLS12-381 G2 public key (compressed)
```
bwbush, line 84: My memory is that other parts of the Ledger use store BLS points uncompressed. I never understood why and I think that Leios should store them compressed.

bwbush (commented):
The RB and vote CDDLs looks good. I noted that the vote CDDL differs from an optimization that the simulators currently use. We'll have to discuss whether that optimization should be memorialized in the specification itself.

bwbush:
For the IB and EB CDDL, we'll need to decide whether a node is allowed to produce multiple IBs in the same slot or multiple EBs in the same pipeline.

- The simulations currently allow this.
- The crypto benchmarks do not.
We've flip-flopped on this a couple of times over the past six months.

Personally, I think the probability of being allowed multiple IBs or EBs is small enough (even for the biggest stakepools) that we should limit it to one. This simplifies the data structures, makes it easier to detect equivocation, etc.

will-break-it:
> Personally, I think the probability of being allowed multiple IBs or EBs is small enough (even for the biggest stakepools) that we should limit it to one. This simplifies the data structures, makes it easier to detect equivocation, etc.

I agree. It would also simplify the network protocol presumably (by disallowing multiple announcements per slot).
I'll stick to a simple CDDL structure when I get to the point. Still looking at the base structure for EBs

docs/cddl/diffs/common/votes-certificates.md
will-break-it: According to [CIP-0381](https://cips.cardano.org/cip/CIP-0381#compressed-vs-decompressed) it seems ok to use compressed and is recommended.
> "even though uncompression is a lot more expensive per point than deserialisation, the size savings due to compression actually outweigh the speed gains due to serialisation because bytes per script are a lot more expensive than ExUnits per script in real terms."

docs/cddl/diffs/common/endorser-blocks.md
```diff
@@ -0,0 +1,83 @@
...
+<sub>[1] [Haskell Simulation EndorserBlock](https://github.com/input-output-hk/ouroboros-leios/blob/main/simulation/src/LeiosProtocol/Common.hs#L160-L171), [2] [Rust Simulation EndorserBlock](https://github.com/input-output-hk/ouroboros-leios/blob/main/sim-rs/sim-core/src/model.rs#L167-L176), [3] [Formal Spec EndorserBlockOSig](https://github.com/input-output-hk/ouroboros-leios-formal-spec/blob/main/formal-spec/Leios/Blocks.agda#L97-L106)</sub>
+
+## Input Block Reference Structure
+
+```cddl
+; References to input blocks within endorser blocks
+ib_reference = [
+  ib_id               : ib_id,                                      ; Hash identifier of the input block
+  slot                : slot_no,                                    ; Slot when IB was created
+  producer            : node_id                                     ; IB producer identifier
```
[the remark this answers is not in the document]
bwbush: I don't think they're strictly needed.

Note that the [Fetch mini protocol spec](https://github.com/input-output-hk/ouroboros-leios/blob/main/docs/technical-report-2.md#fetch-mini-protocol) is silent on the structure of IB references.

docs/cddl/diffs/common/input-blocks.md
```diff
@@ -0,0 +1,58 @@
...
+   [ ib_header_body       : ib_header_body
+   , body_signature       : kes_signature
+   ]
+ 
+ ib_header_body =
+   [ slot                 : slot_no                                  ; Slot when IB was created
+   , producer_id          : pool_id                                  ; Block producer identifier
+   , vrf_proof            : vrf_cert                                 ; VRF proof of eligibility to produce IB
+   , block_body_hash      : hash32                                   ; Hash of the block body
+   , ? ranking_block_ref  : hash32                                   ; Reference to ranking block for ledger state
```
bwbush, line 39: Why is this optional? And have we definitely decided that IBs reference RBs. (We had a lot of discussion in Edinburgh about IBs referencing EBs etc.) I do think that IBs should reference RBs, and that should not be optional, but perhaps we haven't sufficiently thought through the implications of forks.

docs/cddl/diffs/full-sharding/input-blocks.md
```diff
@@ -0,0 +1,39 @@
...
+   ]
+```
+<sub>[1] [Common CDDL](../common/input-blocks.md#block-structure), [2] [Rust simulation](https://github.com/input-output-hk/ouroboros-leios/blob/main/sim-rs/sim-core/src/model.rs#L130)</sub>
+
+## Shard Assignment
+
+```cddl
+; Shard identifier (16-bit allows up to 65,536 shards)
+; Current simulation configs use 1-900 shards
+shard_id = uint .size 2
```
bwbush, line 24: I think this is sufficiently sized.

bwbush (commented):
The block CDDL looks good, but I had a couple of questions and comments.

docs/cddl/diffs/common/votes-certificates.md
```diff
@@ -0,0 +1,141 @@
...
+```
+<sub>[1] [Sig](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/key.rs#L100), [2] [PubKey](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/key.rs#L62), [3] [PoP](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/key.rs#L139-L143), [4] [Eid](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/primitive.rs#L76), [5] [PersistentId](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/registry.rs#L14), [6] [PoolKeyhash](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/primitive.rs#L14), [7] [EbHash](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/primitive.rs#L117), [8] [KesSig](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/src/primitive.rs#L170)</sub>
+
+## Committee Selection
+
+The Fait Accompli algorithm determines persistent voters for each epoch, while local sortition selects non-persistent voters for each election.
```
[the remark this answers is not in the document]
will-break-it: You're right, they don't get transmitted. I'll remove them.

Issue #393, opened 2025-06-09
Haskell simulator crashes on pseudo-mainnet


bwbush:
The Haskell simulator fails due to an assertion failure when running the pseudo-mainnet.

```console
. . .
time reached: 1s
time reached: 1s
time reached: 1.017352999999s
time reached: 1.076184s
time reached: 1.133822s
time reached: 1.202803999999s
time reached: 1.273403s
time reached: 1.34637s
time reached: 1.459996s

ols: Assertion failed
CallStack (from HasCallStack):
  assert, called at src/Chan/TCP.hs:154:19 in ouroboros-leios-sim-0.1.0.0-inplace:Chan.TCP
```

**Steps to reproduce**

1. Configuration: [config.yaml](https://github.com/input-output-hk/ouroboros-leios/blob/a29d0e23cc7be842c010d0d5ce3982e77d6eba70/analysis/sims/2025w23/ibs/haskell-0.5/config.yaml)
2. Network: [network.yaml](https://github.com/input-output-hk/ouroboros-leios/blob/057f58dde4f36ffb3587b5e4ddd48af7646d0aaf/data/simulation/pseudo-mainnet/topology-v1.yaml)
3. Commit: f9990d506e946981187f1f5d50132804493bad1c
4. Command:

```bash
sim leios \
  --leios-config-file config.yaml \
  --topology-file network.yaml \
  --shared-log-format JSON \
  --conformance-events \
  --output-file sim.log \
  --output-seconds 300
```

bwbush:
CC: @nfrisby

nfrisby:
Doh; I had aborted at 40 minutes 🤦

nfrisby:
For reference, this is the assertion that fails and some relevant definitions.

https://github.com/input-output-hk/ouroboros-leios/blob/245de56d037c76ee611495d40a651fd4327bbef5/simulation/src/Chan/TCP.hs#L137C1-L154C53

https://github.com/input-output-hk/ouroboros-leios/blob/245de56d037c76ee611495d40a651fd4327bbef5/simulation/src/Chan/TCP.hs#L176-L185

https://github.com/input-output-hk/ouroboros-leios/blob/245de56d037c76ee611495d40a651fd4327bbef5/simulation/src/ModelTCP.hs#L88-L99

Thus, this assertion fails when a node has not sent anything for a duration X (which is at least 1 second) but still hasn't received all ACKs for messages it sent earlier.

[...]

PR #433, opened 2025-06-26
Avoid floating point false alarms in the TCP model's missing ACK assertion

nfrisby:
Fixes https://github.com/input-output-hk/ouroboros-leios/issues/393.

### DIFF
--- a/simulation/src/Chan/TCP.hs
+++ b/simulation/src/Chan/TCP.hs
@@ -143,15 +143,15 @@ transport tracer tcpprops sendbuf recvbuf = do
     -- if the connection was idle too long, reset the window size
     let tcpstate' :: TcpState
         tcpstate'
-          | now' `diffTime` now <= tcpIdleResetTime tcpprops =
-              tcpstate
-          | otherwise =
+          | tcpIdleResetTime tcpprops now <= now' =
               let allAcksArrived =
                     PQ.foldrWithKeyU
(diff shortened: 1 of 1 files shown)


nfrisby:
@bwbush this runs out of memory even on my 32GB machine, so I can't test my theory.

I've pushed a couple commits so you can.

- First commit skips the "triangle inequality check", which was a slow distraction for this Issue.
- Second commit rearranges the arithmetic/inequality in case it's a false alarm caused by floating point issues.
- Second commit also now prints out the offending numbers for future debugging.

bwbush (approved):
This executes correctly at 327ba1c5117c72d79148e0151ace9fcaba428438 far past the point at which the assertion failed in issue #393, but using the same configuration and network topology.

nfrisby:
Great. I'll tidy it up and merge it today.

nfrisby:
@bwbush FYI this PR also adds the `--skip-triangle-inequality-check` to the `sim` subcommand, since its time complexity is painful with large topologies and it's not particularly critical. (And apparently the real world data doesn't satisfy it.)

nfrisby:
@bwbush FYI this PR also adds the `--skip-triangle-inequality-check` to the `sim` subcommand, since its time complexity is painful with large topologies and it's not particularly critical. (And apparently the real world data doesn't satisfy it.)

PR #438, opened 2025-07-01
Count total_leios_bytes differently

ch1bo:
This seems to be the only way to get a total sum of things that will be stored for FullWithoutIbs and FullWithTxReferences variants.

~~Open point: this will cound transaction bytes double for the other variants and worsen theire spatial efficiency while we would not store txs double of course.~~


SupernaviX:
> Open point: this will cound transaction bytes double for the other variants and worsen theire spatial efficiency while we would not store txs double of course.

@ch1bo good callout, I made this logic variant-specific to cope with that.

PR #452, opened 2025-07-10
Create a dedicated threat model for Leios

ch1bo:
This is a first sketch for a threat model as a dedicated artifact. We have another similar collection of threats here

https://github.com/input-output-hk/ouroboros-leios/blob/main/docs/technical-report-1.md#threat-model

In this document, we follow an established process of threat modeling by also describing a system overview, assets to protect and go more into detail detail on threats, prerequisites, attack vector, cost and impact.

This should be updated with more detail / reference the concrete protocol variant we propose in the CIP and accordingly with assets, threats and mitigations.

### DIFF
--- a/docs/threat-model.md
+++ b/docs/threat-model.md
@@ -113,7 +113,7 @@ Notable threats to the system that could impact assets.
 
 ### Network-Level Threats
 
-#### T1: Mempool Partitioning Attack
+#### T1: Mempool Partitioning
 **Description**: Attacker deliberately partitions the mempools of block producing nodes by submitting conflicting transactions (spending the same inputs) to different network segments, creating inconsistent views of valid transactions across the network.
 
 **Prerequisites**:
@@ -133,13 +133,11 @@ Notable threats to the system that could impact assets.
 **Cost**: MEDIUM-HIGH - Requires significant network infrastructure, multiple nodes, and sustained coordination
 
 **Impact**:
-- **Immediate**: Different SPOs create conflicting EBs, causing vote splits and potential certification failures
-- **Throughput**: Temporary reduction when EBs fail certification, though system recovers in subsequent stages
-- **Resource Waste**: Significant network bandwidth and compute resources wasted on processing, propagating, and voting on conflicting EBs that cannot both be certified
-- **Operational**: SPOs expend resources (CPU, bandwidth) on futile EB creation and voting processes
+- **Throughput**: Different SPOs create conflicting EBs, causing vote splits and potential certification failures. This leads to throughput reduction when EBs fail certification, though system recovers in subsequent stages
+- **Resources**: SPO's network bandwidth and compute resources wasted on processing, propagating, and voting on conflicting EBs that cannot all be certified
 - **Trust**: Demonstrates network manipulation capability, though doesn't break core transaction guarantees
 
-**Assets Affected**: High Throughput, Operational Sustainability, Blockchain Safety
+**Assets Affected**: High Throughput, Operational Sustainability
 
 #### T2: Eclipse Attack on Voting Nodes
 **Description**: Attacker isolates top voting nodes to manipulate vote collection by controlling their network connections and information flow.
@@ -414,11 +412,11 @@ Notable threats to the system that could impact assets.
 **Control type**: Preventive + Detective
 
 **Implementation**:
-- Fair transaction diffusion across peer connections
+- Redundant downstream peer connections and selection similar to upstream
 - Peer connection churn for nodes serving non-chain transactions repeatedly
-- Network topology monitoring for partition detection
-- Redundant peer connections for SPOs
+- Fair transaction diffusion across peer connections
 - Strict limits on perpetual storage (no conflicting tx storage)
+- Network topology monitoring for partition detection
 
 **Validation**: Simulation testing with network partitions
 
@@ -434,12 +432,10 @@ Notable threats to the system that could impact assets.
 **Control type**: Preventive + Detective
 
 **Implementation**:
-- Diversified peer connections for all nodes
-- Network monitoring and anomaly detection
-- BGP route monitoring for critical nodes
-- Automated peer rotation mechanisms
+- Redundant and tiered upstream peer connections
 - Detect and punish censoring across mini protocols
 - Apply protection to all critical path network protocols (vote diffusion, EB diffusion, block fetch)
+- Network monitoring and anomaly detection
 
 **Validation**: Penetration testing and network analysis
 
--- a/docs/threat-model.md
+++ b/docs/threat-model.md
@@ -1,0 +1,553 @@
+---
+Authors: Sebastian Nagel
+Document Version: 1.0
+Last Updated: 2025-07-10
+Next Review: 2025-09-10
+---
+# Leios Consensus Upgrade - Threat Model
+
+A threat model for the Leios consensus change for Cardano. This was created by roughly following the [OWASP threat modeling process](https://owasp.org/www-community/Threat_Modeling_Process).
+
+## 1. System Overview
+
+> [!NOTE]
+> The described system here is the heavily simplified EB-only variant of Leios. Whenever we update this, reflect on existing assets, threats and mitigations, as well as add new ones accordingly.
+
+### 1.1 Description
+Leios is an overlay protocol on top of Ouroboros Praos that enhances transaction throughput by introducing Endorser Blocks (EBs) alongside regular Praos blocks (Ranking Blocks - RBs). The system maintains backward compatibility at the client interface while introducing new responsibilities for stake pools.
+
+### 1.2 Key Components
+
+#### Core Components
+- **Endorser Blocks (EBs)**: New block type that references transactions for inclusion, up to 640 kB (~20k transaction references)
+- **Ranking Blocks (RBs)**: Standard Praos blocks enhanced with EB certificates
+- **Vote Certificates**: Aggregated stake-weighted votes on EB validity and transaction availability
+- **VRF-based EB Eligibility**: Separate lottery system for EB creation rights
+
+#### Network Protocols
+- **EB Announcement Protocol**: Gossip protocol for EB existence
+- **EB Fetch Protocol**: Pull-based protocol for retrieving EBs
+- **Vote Diffusion Protocol**: Pull-based protocol for vote propagation
+- **Transaction Fetch Protocol**: Enhanced protocol for retrieving endorsed transactions
+
+#### Participants
+- **Stake Pool Operators**: Enhanced with EB creation and voting responsibilities
+- **Top 500 Voting Nodes**: Subset of stake pools by stake size, responsible for voting on EBs
+- **Regular Nodes**: Participate in transaction and block propagation
+- **Clients**: Maintain backward compatibility, unaware of Leios mechanics
+
+### 1.3 System Flow
+1. Stake pools create EBs based on VRF eligibility (parameterizable stage length)
+2. EBs are announced and propagated through the network
+3. Top 500 nodes by stake vote on EB validity and transaction availability
+4. If 75% of voting stake approves, a certificate is created
+5. Certificates are included in the next available RB (every ~20 seconds)
+6. Missing transactions are fetched on-demand when EBs are processed
+
+## 2. Assets to Protect
+
+For each asset we define what could be impacted in respect to its Confidentiality, Integrity, Availability; i.e. the [CIA Triad](https://www.splunk.com/en_us/blog/learn/cia-triad-confidentiality-integrity-availability.html)
+
+### 2.1 Blockchain Safety
+**Description**: The fundamental guarantee that all honest nodes agree on the blockchain history and no conflicting valid chains exist.
+
+**CIA Impact:**
+- **Confidentiality**: Not applicable - blockchain is public
+- **Integrity**: CRITICAL - Chain splits or conflicting histories would break consensus
+- **Availability**: HIGH - Safety failures could halt the network indefinitely
+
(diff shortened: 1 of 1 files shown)


docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+
+### 1.1 Description
+Leios is an overlay protocol on top of Ouroboros Praos that enhances transaction throughput by introducing Endorser Blocks (EBs) alongside regular Praos blocks (Ranking Blocks - RBs). The system maintains backward compatibility at the client interface while introducing new responsibilities for stake pools.
+
+### 1.2 Key Components
+
+#### Core Components
+- **Endorser Blocks (EBs)**: New block type that references transactions for inclusion, up to 640 kB (~20k transaction references)
+- **Ranking Blocks (RBs)**: Standard Praos blocks enhanced with EB certificates
+- **Vote Certificates**: Aggregated stake-weighted votes on EB validity and transaction availability
```
bwbush, line 24: In the first line these are called "EB certificates" but in the second line they are called "vote certificates". Perhaps we should use a single name for this component.

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+#### Network Protocols
+- **EB Announcement Protocol**: Gossip protocol for EB existence
+- **EB Fetch Protocol**: Pull-based protocol for retrieving EBs
+- **Vote Diffusion Protocol**: Pull-based protocol for vote propagation
+- **Transaction Fetch Protocol**: Enhanced protocol for retrieving endorsed transactions
+
+#### Participants
+- **Stake Pool Operators**: Enhanced with EB creation and voting responsibilities
+- **Top 500 Voting Nodes**: Subset of stake pools by stake size, responsible for voting on EBs
+- **Regular Nodes**: Participate in transaction and block propagation
```
bwbush, line 36: Should this be split into relays vs block producers not in the 500.

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+
+### 3.1 Network-Level Threats
+
+#### T1: Mempool Partitioning Attack
+**Description**: Attacker deliberately partitions the mempools of block producing nodes by submitting conflicting transactions (spending the same inputs) to different network segments, creating inconsistent views of valid transactions across the network.
+
+**Prerequisites**:
+- Control over multiple network positions to segment the peer-to-peer network
+- Ability to create valid but conflicting transactions (same inputs, different outputs)
+- Knowledge of network topology and SPO peer relationships
```
bwbush, line 121:
```suggestion
- Discovery of network topology and SPO peer relationships
```

The point is that anyone can map the network topology. The work "knowledge" might erroneously imply access to non-public information.

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+
+**Attack Vector**:
+1. Attacker maps network topology and identifies cluster boundaries
+2. Creates conflicting transaction pairs spending identical UTXOs
+3. Submits Transaction A to Network Segment 1, Transaction B to Network Segment 2
+4. Uses network position control (BGP, routing, eclipse techniques) to prevent cross-segment propagation
+5. SPOs in different segments create EBs endorsing different conflicting transactions
+6. Voting nodes must choose between conflicting EBs, potentially causing certification failures
+
+**Cost**: MEDIUM-HIGH - Requires significant network infrastructure, multiple nodes, and sustained coordination
```
bwbush, line 132:
```suggestion
**Cost**: MEDIUM - Requires significant network infrastructure, multiple nodes, and sustained coordination
```
This doesn't require deploying many nodes if they are nodes custom-modified for this attack. The attack would require more nodes if vanilla ones are used.

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+
+**Impact**:
+- **Limited Effect**: Multiple certificates can exist but only RB inclusion affects chain selection
+- **Resource Waste**: Additional bandwidth and processing for redundant certificates
+- **No Safety Violation**: Chain safety depends on RB inclusion, not certificate existence
+- **Detection Difficulty**: Voters may have plausible deniability if conflicting EBs aren't clearly marked as competing
+
+**Assets Affected**: Operational Sustainability
+
+#### T6: VRF Manipulation
```
bwbush, line 243: We might want to cross-reference the anti-grinding CPS and CIP.

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+#### T9: Transaction Front-Running
+**Description**: EB producers observe profitable transactions and reorder or insert their own transactions to extract value before the original transaction executes.
+
+**Prerequisites**:
+- EB creation eligibility
+- MEV (Maximal Extractable Value) opportunities in transaction sets
+- Knowledge of transaction dependencies and profitable patterns
+
+**Attack Vector**:
+1. Monitor mempool for profitable transaction patterns
```
bwbush, line 323: Is this really different from the situation with Praos?

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+- Apply protection to all critical path network protocols (vote diffusion, EB diffusion, block fetch)
+
+**Validation**: Penetration testing and network analysis
+**Cost**: Medium - Monitoring infrastructure and operational procedures
+**Threats Addressed**: T2
+
+#### M3: Vote Flooding Protection
+**Decision**: MITIGATE
+**Control Type**: Preventive
+**Implementation**:
```
bwbush, line 443: We also need to clarify the protocol rules.

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+**Impact**:
+- **Throughput**: Reduced transaction processing capacity when EBs are withheld
+- **Censorship**: Potential to delay specific transactions if coordinated
+- **Temporary**: System recovers with next EB opportunity or alternative producers
+- **Limited**: Cannot permanently block transactions due to multiple eligibility opportunities
+
+**Assets Affected**: High Throughput, Decentralization Properties
+
+#### T5: Double Voting
+**Description**: Top 500 voting nodes vote on multiple conflicting EBs that reference the same transactions or create incompatible blockchain states.
```
bwbush, line 220: Doesn't the ledger-update protocol sort this out? What, precisely, are "conflicting EBs" aside from EBs that contain mutually incompatible transactions?

bwbush (changes requested):
The vote/certificate assumptions here ("top 500 nodes" etc.) clash with the previously reviewed specification. Is this document proposing a different approach to vote and certificate cryptography? If we we need a detailed specification that the cryptography team can study.

At some point we might want to align terminology between this document and the [Notes on Attack Surface](https://github.com/input-output-hk/ouroboros-leios/blob/main/docs/technical-report-2.md#notes-on-the-leios-attack-surface) in the 2nd Tech report before we move any of this material into the Leios CIP itself.

A lot of the attack vectors are means to get conflicting txs in the EBs, so that resource are wasted. It might be helpful to discuss that in more detail, and centrally, so that individual threat and mitigation sections can reference that.

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+- **Trust**: Demonstrates network manipulation capability, though doesn't break core transaction guarantees
+
+**Assets Affected**: High Throughput, Operational Sustainability, Blockchain Safety
+
+#### T2: Eclipse Attack on Voting Nodes
+**Description**: Attacker isolates top voting nodes to manipulate vote collection by controlling their network connections and information flow.
+
+**Prerequisites**:
+- Control over significant network infrastructure (BGP routes, ISPs, or direct node connections)
+- Knowledge of top 500 voting nodes' network addresses and topology
```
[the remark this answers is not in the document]
ch1bo: Can you contribute a reference to this on the system overview, threat or mitigation of this?

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+**Description**: Top 500 voting nodes vote on multiple conflicting EBs that reference the same transactions or create incompatible blockchain states.
+
+**Prerequisites**:
+- Membership in top 500 voting nodes by stake
+- Access to multiple conflicting EBs in the same voting period
+- Malicious intent or compromised voting infrastructure
+
+**Attack Vector**:
+1. Receive or create multiple conflicting EBs for the same stage
+2. Submit votes for multiple conflicting EBs instead of choosing one
```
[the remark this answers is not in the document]
ch1bo: Further detail on the system scope could clarify this for sure. The threat exists even in any case, but we may choose to not discuss it explicitly if we can avoid (control type of a mitigation) it by design.. or keep it nonetheless

docs/threat-model.md
ch1bo: No, but that does not make it less of a threat and I wanted to cover a range of threats which had been in discussion lately. What's your suggestion here?

docs/threat-model.md
```diff
@@ -0,0 +1,553 @@
...
+**Threats Addressed**: T8
+
+### 5.2 Consensus-Level Controls
+
+#### M5: EB Withholding Mitigation
+**Decision**: MITIGATE
+**Control Type**: Preventive
+**Implementation**:
+- Parameterize EB opportunities and sizes for adversarial stake assumptions
+- Example: Assume 30% adversarial stake, produce 2 EBs per stage on average
```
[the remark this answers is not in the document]
ch1bo: What do you mean? A threat that our assumed adversarial stake is wrong?

docs/threat-model.md
ch1bo: The description is imprecise here, I'll fix it up on monday

docs/threat-model.md
```diff
@@ -0,0 +1,590 @@
...
+#### Core Components
+- **Ranking Block (RB)**: Standard Praos block enhanced with a Leios certificate
+- **Endorser Block (EB)**: New block type that references transactions for inclusion. May be substantially bigger than Praos blocks (currently on mainnet ~88 kB) with sizes around ~640 kB for 20k transaction references
+- **Leios Certificate**: Cryptographic proof about aggregated stake-weighted votes on EB validity and transaction availability
+- **EB Lottery**: Separate (to Praos) VRF lottery for EB creation rights
+
+#### Network Protocols
+- **EB Announcement Protocol**: Gossip protocol about EB existence
+- **EB Fetch Protocol**: Pull-based protocol for retrieving EBs
+- **Transaction Fetch Protocol**: Pull-based protocol for retrieving endorsed transactions
```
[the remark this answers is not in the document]
ch1bo: The first

docs/threat-model.md
```diff
@@ -0,0 +1,591 @@
...
+- Control over multiple network positions to segment the peer-to-peer network
+- Ability to create valid but conflicting transactions (same inputs, different outputs)
+- Discovery of network topology and SPO peer relationships
+- Timing coordination to submit conflicts before natural mempool synchronization
+
+**Attack Vector**:
+1. Attacker maps network topology and identifies cluster boundaries
+2. Creates conflicting transaction pairs spending identical UTXOs
+3. Submits Transaction A to Network Segment 1, Transaction B to Network Segment 2
+4. Uses network position control (BGP, routing, eclipse techniques) to prevent cross-segment propagation
```
WhatisRT, line 134: I don't think this is necessary. Even if all EB producers know about both transactions, how would you force them to all pick the same one? With random selection you'd get a 50% probability that both are selected so the expected value would be 1.5 transactions for the price of one. Now you can improve that in various ways, segmenting the network being one of them, but you can also control some stake (to make EBs that you know will conflict) or if you know about the EB schedule you can probably optimize who you send your transactions to. But also, depending on the details of an attack you might not even need to increase the probability of the conflict in the first pace.

ch1bo: > I don't think this is necessary.

I think it is a prerequisite in order to get conflicting transactions into the mempools of concurrently producing nodes. One would need to defeat the natural (fair?) tx diffusion happening between nodes. I should make that more clear in the attack vector here.

> can also control some stake (to make EBs that you know will conflict)

I'd characterize this as a different attack due to the requirements and attack vector being quite different. This one is really about a network-attacker that does not control stake. Do you feel this would be much different to also T4 or T8 to warrant a dedicated entry? Would this "stake-based deliberate conflict" attack result in different impacts onto assets in the given system than the ones mentioned?

> depending on the details of an attack you might not even need to increase the probability of the conflict in the first pace.

The attack here is about creating conflicting views, i.e. one or more transactions conflict between concurrently created blocks.

docs/threat-model.md
```diff
@@ -0,0 +1,637 @@
...
+**Cost**: HIGH - Requires substantial capital investment in stake acquisition
+
+**Impact**:
+- **Voting Bias**: Disproportionate influence over EB certification
+- **Eligibility Manipulation**: Increased EB creation opportunities
+- **Centralization**: Concentration of power despite appearing decentralized
+
+**Assets Affected**: Decentralization
+
+#### T?: Honey Pot Contract
```
ch1bo, line 294: @WhatisRT Can you check whether I summarized your attack correctly?

docs/threat-model.md
bwbush: For this document we don't need to change anything, but when representation Leios we might want to clearly distinguish three categories of threats:

1. Exists in Praos and similarly exists in Leios.
2. Exists in Praos but more serious in Leios.
3. Does not exist in Praos but does exist in Leios.

docs/threat-model.md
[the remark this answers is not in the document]
bwbush: I guess I'd like clarity about whether Leios should be robust against a 49% adversary, like Praos is theoretically. Or instead maybe we're aiming for robustness against a weaker adversary, perhaps 30%.

bwbush (approved):
My comments were adequately addressed. It looks ready now. Thanks!

PR #455, opened 2025-07-12
Linear Leios

SupernaviX:
See `LINEAR_LEIOS.md` for usage.

### DIFF
--- a/sim-rs/LINEAR_LEIOS.md
+++ b/sim-rs/LINEAR_LEIOS.md
@@ -10,9 +10,9 @@ Whenever a node creates an RB, it also creates an EB. The RB header contains a r
 
 RB headers are diffused separately from bodies. When a node receives an RB header, it checks whether that RB should be the new head of its chain. If so, it will request the RB body and the referenced EB (from the first peer which announces them).
 
-When a node receives an RB body, it immediately removes all referenced/conflicting transactions from its mempool. If the RB has an EB certificate, it also removes that EB’s transactions from its mempool. 
+When a node receives an RB body, it immediately removes all referenced/conflicting transactions from its mempool. If the RB has an EB certificate, it also removes that EB’s transactions from its mempool (note that this should be redundant based on the procedure below).
 
-When a node receives an EB body, it immediately runs a VRF lottery, and if successful, transmits votes for that EB. If the EB has been certified after L_vote + L_diff slots have passed, the node removes all of its transactions from the mempool (under the assumption that the EB will make it on-chain).
+When a node receives an EB body, it immediately runs a VRF lottery to decide how many times it can vote for that EB; if it has any votes, it will transmit them to all peers. If the EB has been certified after L_vote + L_diff slots have passed, the node removes all of its transactions from the mempool (under the assumption that the EB will make it on-chain).
 
 ## New parameters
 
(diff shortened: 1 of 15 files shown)


sim-rs/LINEAR_LEIOS.md
```diff
@@ -0,0 +1,30 @@
...
+
+The log file schema is currently identical to every other variant (though `pipeline` is always 0).
+
+## Description
+
+Whenever a node creates an RB, it also creates an EB. The RB header contains a reference to this new EB. If the RB producer has a certificate for the parent RB’s EB, it will include that certificate in the RB body.
+
+RB headers are diffused separately from bodies. When a node receives an RB header, it checks whether that RB should be the new head of its chain. If so, it will request the RB body and the referenced EB (from the first peer which announces them).
+
+When a node receives an RB body, it immediately removes all referenced/conflicting transactions from its mempool. If the RB has an EB certificate, it also removes that EB’s transactions from its mempool.
```
nfrisby, line 13: > `If the RB has an EB certificate, it also removes that EB’s transactions from its mempool.`

When the node issues an RB, which txs does it include? If the above rule is the only way EBs evict txs from the mempool, then I'm suspicious that the RB duplicates a prefix of its parent's EB.

sim-rs/LINEAR_LEIOS.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Description
+
+Whenever a node creates an RB, it also creates an EB. The RB header contains a reference to this new EB. If the RB producer has a certificate for the parent RB’s EB, it will include that certificate in the RB body.
+
+RB headers are diffused separately from bodies. When a node receives an RB header, it checks whether that RB should be the new head of its chain. If so, it will request the RB body and the referenced EB (from the first peer which announces them).
+
+When a node receives an RB body, it immediately removes all referenced/conflicting transactions from its mempool. If the RB has an EB certificate, it also removes that EB’s transactions from its mempool. 
+
+When a node receives an EB body, it immediately runs a VRF lottery, and if successful, transmits votes for that EB. If the EB has been certified after L_vote + L_diff slots have passed, the node removes all of its transactions from the mempool (under the assumption that the EB will make it on-chain).
```
nfrisby, line 15: > `it immediately runs a VRF lottery`

Could you elaborate on this part? Maybe it's merely checking whether the voting leader schedule elects it for the slot of the EB?

sim-rs/LINEAR_LEIOS.md
SupernaviX: @nfrisby good callout, this mempool pruning is actually redundant. If the parent EB was produced >= `L_vote + L_diff` slots ago and certified in time, the node would already have evicted its TXs from the mempool..

sim-rs/LINEAR_LEIOS.md
SupernaviX: When a node sees an EB, it runs a (simulated) VRF lottery to decide how many votes it has the right to produce. I don't know how VRF lotteries work for real, but if they have to be tied to a slot then we could use the slot of the EB for that (and I guess a node just has the right to vote for all EBs produced in a slot if there are slot battles).

Totally possible that I'm abusing the term or concept of VRF lottery here. my cryptographic background is pretty week.

Issue #453, opened 2025-07-10
Haskell simulator's multiplexed TCP channel model has infelicitous latency


nfrisby:
This is the code that constructs the channels between the nodes of our simulation.

https://github.com/input-output-hk/ouroboros-leios/blob/acb6b1bc23d48e421ed181181d46d27761fb0e20/simulation/src/LeiosProtocol/Short/SimP2P.hs#L56-L60

And this is the code of the called function.

https://github.com/input-output-hk/ouroboros-leios/blob/acb6b1bc23d48e421ed181181d46d27761fb0e20/simulation/src/Chan.hs#L62-L90

In the last two cases are the "realistic" ones, since they involve TCP, and only the last is particularly realistic since it uses both TCP and the multiplexed connection.

- If `_mux` is `False`, then each mini protocol in the bundle gets its own instance of the connection _with the given latency and bandwidth_. This could only be realistic if the bandwidth were infinite or if the given bandwidth were _statically_ pre-divided amongst the mini protocols.
- If `_mux` is `True`, then there is one TCP connection that carries all of the mini protocols' messages. However, it only carries one message at a time. This is unrealistic because it will have worse latency for messages sent while a message is in flight than does the real node's implementation, because the real node slices each message into parts and sends them round-robin among the mini protocols. Thus a mini protocol message's latency is only increased by a small multiple of "partSize/bandwidth" and its bandwidth is a fair proportion of however many mini protocols simultaneously have message slices to send.

This Issue is to improve the "message bundle" abstraction so that the TCP+mux model utilizes the topology link's latency and bandwidth as well as the real node does.

The downside is that the higher fidelity will likely decrease simulation performance.

SupernaviX:
If it helps, the rust bandwidth implementation and tests [are here](https://github.com/input-output-hk/ouroboros-leios/blob/main/sim-rs/sim-core/src/network/connection.rs). Unfortunately, it's not modeling anything at the packet level (which you probably need to do for simulating TCP effects accurately). It's just divvying up the bandwidth across mini-protocols, and pretending they're all in flight at once

...Now that I think about it, modeling this as packets would probably have been easier to implement.

PR #466, opened 2025-07-21
Linear Leios without txs

nfrisby:
This PR adds Linear Leios to the Haskell simulator.


bwbush:
The EB validation takes five seconds--or am I reading this log wrong? That delays the time from vote generation to the start of diffusion. Wouldn't the EB be validated before it is voted upon?

```json
{"message":{"id":"77--1","pipeline":0,"producer":"node-79","size_bytes":105,"slot":53,"type":"VTBundleGenerated","votes":{"77-0":3}},"time_s":56.000163}
{"message":{"id":"77-0","recipient":"node-20","type":"EBReceived"},"time_s":63.096319}
{"message":{"cpu_time_s":5.049999,"id":"77-0","node":"node-20","task_type":"ValEB","type":"Cpu"},"time_s":63.096319}
{"message":{"id":"77--1","msg_size_bytes":138,"recipient":"node-20","sender":"node-79","sending_s":1.35e-4,"type":"VTBundleSent"},"time_s":63.130592}
```

nfrisby:
In the example I'm looking at (EB 97-0, VB 97--1), I suspect that the EB takes 9-10 seconds to arrive to a direct neighbor node 72, and head-of-line blocking on the channel prevents the VB from arriving before that.

The EB is generated by node 97 at 493.130 and it arrives at node 72 (a direct neighbor, I think) at 503.000. The VB is generated by node 97 at 496.000 at arrives at 503.038. (edited)  And that blockage would then replicate at every hop. So I think https://github.com/input-output-hk/ouroboros-leios/issues/453 plausibly explains the delay we're seeing.

-----

The "sent" events' timestamps are also delayed for this same reason, which might be surprising. I'm not sure, but I strongly suspect the "sent" events are only emitted when the Relay mini protocol actually sends the body. The mini protocol, though, exchanges a couple messages before sending the body, and those messages are also blocked behind the EB in the channel.

nfrisby:
@bwbush In the example you gave, the vote arises before the EB is validated because of the exceptional case of the voting node being the one that generated the EB.

```
$ zcat sim.log.gz | grep -e node-79 | grep -e '77-0'
{"message":{"endorser_blocks":[],"id":"77-0","input_blocks":[],"pipeline":0,"producer":"node-79","size_bytes":10000304,"slot":53,"type":"EBGenerated"},"time_s":53.13}
{"message":{"id":"77--1","pipeline":0,"producer":"node-79","size_bytes":105,"slot":53,"type":"VTBundleGenerated","votes":{"77-0":3}},"time_s":56.000163}
```

I don't know why "node-79" has a numeric id of 77 🤷 

whereas "node-97" has a numeric id of 99 😵‍💫 

```
$ zcat sim.log.gz | grep -e node-97 | grep -e '97-0'
{"message":{"endorser_blocks":[],"id":"97-0","input_blocks":[],"pipeline":0,"producer":"node-97","size_bytes":10000304,"slot":493,"type":"EBGenerated"},"time_s":493.129999}
{"message":{"id":"97--1","pipeline":0,"producer":"node-97","size_bytes":105,"slot":493,"type":"VTBundleGenerated","votes":{"97-0":5}},"time_s":496.000163}
```

nfrisby:
I pushed up the `[Messages are now interleaved when multiplex-mini-protocols: true](https://github.com/input-output-hk/ouroboros-leios/pull/466/commits/33078f51f7489cbda3728a328977492dd66923b9)` commit, and it seems to have addressed the unexpected 9 second delay. We're now seeing a 3 second delay, which is due to implementing the spec's $3\Delta_hdr$ moratorium on voting (if I'm interpreting the coordinates of the plot correctly).

https://github.com/input-output-hk/ouroboros-leios/blob/main/analysis/sims/2025w32b/analysis.ipynb

image

Thanks Brian for running the confirmation sim so quickly.

PR #486, opened 2025-08-06
Update threat model for Linear Leios

ch1bo:
Updated scope to fit the latest design we have in the CIP draft. Updated several threats in there. I'm not 100% happy with the level of detail and its honestly a bit hard to maintain (also the diffs are horrible with re-ordering things). Especially as there are very good write-ups of attacks already.

Some key changes / observations:
- Linear leios does not allow us to use mitigate withholding and front-running by over-parameterization (`M5`)
- Linear leios avoids conflicting transactions by design. This is covered by `M11`, but I'm not sure whether things avoided by design should be actual mitigations?

@will-break-it Feel free to use it or leave it out for the CIP rationale/appendix.

### DIFF
--- a/docs/threat-model.md
+++ b/docs/threat-model.md
@@ -8,7 +8,7 @@ Next Review: 2025-10-05
 
 # Leios Consensus Upgrade - Threat Model
 
-A threat model for the Linear Leios consensus change for Cardano. This reflects the simplified "Linear Leios" variant described in the CIP draft, which eliminates Input Blocks (IBs) and produces Endorser Blocks (EBs) alongside Ranking Blocks (RBs) by the same block producer.
+A threat model for the Leios consensus change for Cardano. This reflects the simplified "Linear Leios" variant described in the CIP draft, which eliminates Input Blocks (IBs) and produces Endorser Blocks (EBs) alongside Ranking Blocks (RBs) by the same block producer.
 
 See also [the threat model section in Leios Technical Report #1](./technical-report-1.md#threat-model) and more [comments on attack surface in Leios Technical Report #2](./technical-report-2.md#notes-on-the-leios-attack-surface).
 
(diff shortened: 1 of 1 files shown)


docs/threat-model.md
```diff
@@ -630,7 +611,25 @@ Notable threats to the system that could impact assets.
...
+
+### M11: No Conflicting Transactions
+**Addressing threats**: T1, T12
+
+**Decision**: MITIGATE
+
+**Control type**: By design
+
+**Implementation**:
+- Protocol design inherently prevents conflicting transactions from reaching the chain
```
bwbush, line 624: Not 100% true any more, because of the need for execution bitmaps.

docs/threat-model.md
```diff
@@ -291,102 +286,56 @@ Notable threats to the system that could impact assets.
 
 **Assets Affected**: Decentralization
 
-#### T?: Honey Pot Contract
-
-**Description**: An attacker deliberately makes ADA available on-chain so anyone races to claim it with the goal of producing many conflicting transactions. This is very similar to T1, but uses cryptocurrency instead of network resources.
+#### T8: Transaction Withholding
```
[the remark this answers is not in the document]
ch1bo: Why would delayed release be more powerful than not release at all?

Admittedly, this thread is very similar to T7 (EB withholding) and we maybe should analyse the delaying too. The difference I can see with a delay attack is (hard to mount) where the delay results in a certificate created (quorum got EB + txs in time), but availability to the remainder of the network is delayed with the goal to interfere with RB chain extension (threatening blockchain safety or liveness)

docs/threat-model.md
ch1bo: True, my PR predates their introduction and I'm not fully on board with those yet :)

PR #498, opened 2025-08-18
DO NOT MERGE WIP CIP section on mini protocols

nfrisby:
This PR is just to share an early draft of a section I'm writing for the CIP.

### DIFF
--- a/NewLinearLeiosMiniProtocols.md
+++ b/NewLinearLeiosMiniProtocols.md
@@ -69,21 +69,23 @@ The following rules would be plausible.
 
 **Block age bounds**.
 LLBs have three diffusion intervals.
-This first interval is analogous to votes but lasts L_recover instead of L_vote.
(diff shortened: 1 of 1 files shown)


nfrisby:
Alright, as of the latest push, only minor TODOs remain. The only non-trivial work I anticipate is pruning/reorganizing.

In other words: this first draft includes at least as much technical detail as we want in the CIP, but maybe more.

NewLinearLeiosMiniProtocols.md
```diff
@@ -0,0 +1,593 @@
+## Mini Protocols
```
nfrisby, line 1: I suspect we should pivot this section to instead be a "demonstration of feasibility". IE we're not necessarily claiming that this is what the wire protocol should be, just that it plausibly could be this without requiring excessive resources.

For example, promise a more concrete follow-up CIP that is intended to pin down exact concrete details.

PR #507, opened 2025-08-21
CIP: Path to active

ch1bo:


docs/cip/README.md
```diff
@@ -2305,50 +2305,46 @@ incentives and conflict resolution trade-offs.
...
+are:
+
+- [x] Simulations to confirm general feasibility using a model of mainnet.
+- [x] Ecosystem impact analysis and establish a threat model.
+- [ ] Coordinate with related activities on other protocol enhancements
+      (Phalanx, Peras, Mithril).
+- [ ] Detailed node-level (as opposed to this protocol-level) technical
+      specification.
+- [ ] Complete ΔQ analysis of new/changed network interactions.
+- [ ] Complete consensus-level formal protocol specification in Agda.
```
bwbush, line 2334: What about network and ledger specs?

ch1bo: I agree about the ledger part (I kind of implied it).
Did we ever "drill down" to the network-level with these agda specs?

PR #521, opened 2025-08-25
CIP Security Argument

will-break-it:


ch1bo (changes requested):
I like the added content of a Praos security argument, I don't like the notion of "protocol phases" as it is yet another narrative dimension and it is not really needed IMO (see also my comments and suggestions on the main PR).

I feel like we can simplify the overall structure by just having three relevant things:
- protocol flow: with narrative steps which mention protocol parameters
- protocol parameters: introduce all parameters and define what they mean + any interaction / relations between them
- protocol security: analysis / argument why this protocol is secure

The security argument heavily relies on an assumption which is used to constrain protocol parameters, nothing more, nothing less. This makes a strong connection between security argument and protocol parameters. So having these two close by / one being part of the other would make sense to me.

ch1bo (approved):
Having the timing constraints in the protocol parameter section makes more sense to me, thanks.

PR #557, opened 2025-09-26
Crypto impact analysis

perturbing:
Hi, 

Here is an initial attempt at the impact analysis of Leios on the cryptography of the node. As an addition to #555 

@curiecrypt @ch1bo

### DIFF
--- a/docs/ImpactAnalysis.md
+++ b/docs/ImpactAnalysis.md
@@ -421,28 +421,28 @@ The security of the votes cast and the certificates that Leios uses to accept EB
 
 This section derives **requirements** for adding BLS signatures to `cardano-base` and sketches **changes** to satisfy them. The scope is limited to cryptographic primitives and their integration into existing classes; vote construction/logic is out of scope. This work should align with [this](https://www.ietf.org/archive/id/draft-irtf-cfrg-bls-signature-05.html) IETF draft.
 
-> Note that with the implementation of [CIP-0381](https://cips.cardano.org/cip/CIP-0381) `cardano-base` already contains basic utility functions needed to create these bindings; the work below is thus expanding on that.
+> Note that with the implementation of [CIP-0381](https://cips.cardano.org/cip/CIP-0381) `cardano-base` already contains basic utility functions needed to create these bindings; the work below is thus expanding on that. The impact of the below requirements thus only extends to [this](https://github.com/IntersectMBO/cardano-base/blob/82e09945726a7650540e0656f01331d09018ac97/cardano-crypto-class/src/Cardano/Crypto/EllipticCurve/BLS12_381/Internal.hs) module and probably [this](https://github.com/IntersectMBO/cardano-base/blob/82e09945726a7650540e0656f01331d09018ac97/cardano-crypto-class/src/Cardano/Crypto/DSIGN/Class.hs) outward facing class.
 
 ## Requirements
 
 ### Functional
 
+- *REQ-BlsTypes*.
+Introduce opaque types for `SecretKey`, `PublicKey`, `Signature`, and `AggSignature` (if needed by consensus).
 - *REQ-BlsKeyGenSecure*.
 Provide secure key generation with strong randomness requirements, resistance to side-channel leakage.
 - *REQ-BlsVariantAbstraction*.
 Support both BLS variants—small public key and small signature—behind a single abstraction. Public APIs are variant-agnostic.
+- *REQ-BlsPoP*.
+Proof-of-Possession creation and verification to mitigate rogue-key attacks.
 - *REQ-BlsSkToPk*.
 Deterministic sk → pk derivation for the chosen variant.
 - *REQ-BlsSignVerify*.
 Signature generation and verification APIs, variant-agnostic and domain-separated (DST supplied by caller). Besides the DST, the interface should also implement a per message augmentation (as the hash to curve function also has in the IETF draft)
-- *REQ-BlsPoP*.
-Proof-of-Possession creation and verification to mitigate rogue-key attacks.
 - *REQ-BlsAggregateSignatures*.
 Aggregate a list of public keys and signatures into one
 - *REQ-BlsBatchVerify*.
 Batch verification API for efficient verification of many `(pk, msg, sig)` messages.
-- *REQ-BlsTypes*.
-Introduce opaque types for `SecretKey`, `PublicKey`, `Signature`, and `AggSignature` (if needed by consensus).
 - *REQ-BlsDSIGNIntegration*.
 Provide a `DSIGN` instance so consensus can use BLS via the existing `DSIGN` class, including aggregation-capable helpers where appropriate.
 - *REQ-BlsSerialisation*.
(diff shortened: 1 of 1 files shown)


docs/ImpactAnalysis.md
```diff
@@ -263,8 +263,51 @@ It is not already clear which new or updated mechanisms/components would mitigat
...
+- *REQ-BlsAggregateSignatures*.
+Aggregate a list of public keys and signatures into one
+- *REQ-BlsBatchVerify*.
+Batch verification API for efficient verification of many `(pk, msg, sig)` messages.
+- *REQ-BlsTypes*.
+Introduce opaque types for `SecretKey`, `PublicKey`, `Signature`, and `AggSignature` (if needed by consensus).
+- *REQ-BlsDSIGNIntegration*.
+Provide a `DSIGN` instance so consensus can use BLS via the existing `DSIGN` class, including aggregation-capable helpers where appropriate.
+- *REQ-BlsSerialisation*.
+Deterministic serialisation: `ToCBOR`/`FromCBOR` and raw-bytes for keys/signatures; strict length/subgroup/infinity checks; canonical compressed encodings as per the Zcash standard for BLS points.
```
ch1bo, line 293: Can you link this standard?

docs/ImpactAnalysis.md
```diff
@@ -263,8 +263,51 @@ It is not already clear which new or updated mechanisms/components would mitigat
...
+- *REQ-BlsConformanceVectors*.
+Add conformance tests using test vectors from the Rust implementation to ensure cross-impl compatibility.
```
ch1bo, line 295: Which implementation; link?

ch1bo (commented):
Had a first look and you seem comfortable working with this way of sketching requirements like @nfrisby did. You are probably still working on it so maybe not required to ask:
- What concrete changes are needed? Do we need to implement anything new or is it just integration of a new library or maybe even new functions from an existing library?
- Maybe for the introduction/giving context: this seems to rely that BLS is the scheme we are going with, could be worthwhile to point out this (reasonable) assumption
- Anything to say about Peras?

docs/ImpactAnalysis.md
perturbing: https://github.com/zcash/librustzcash/blob/6e0364cd42a2b3d2b958a54771ef51a8db79dd29/pairing/src/bls12_381/README.md#serialization

This is already what the current `cardano-base` implementation uses for point compression.

Ill add the link

ch1bo:
> I tried, but please let me know if something is not in line with this way of formalizing the requirements. I just followed suit ;)

I have not tried it myself .. let's see if it sticks :)

curiecrypt:
I just finished reading CIP in a bit more detail. I also read the impact analysis draft of yours @perturbing and I think it seems pretty mature. Although I might have some additions after discussing with you.
To respond to your questions @ch1bo:
- BLS scheme seems the most reasonable one. There is nothing wrong with pointing that out by giving the reasoning IMO.
-  I believe [blst](https://crates.io/crates/blst) library covers all the cryptographic functionality we need. However, as @perturbing stated in the document, we are going to introduce opaque types and wrappers around the `blst` library. Also, 
even if it is not possible to use `mithril-stm` as is, we can make use of a lot of coding practice from the library.
- A specific section could be added to point out the parts related/unrelated to Peras.

docs/ImpactAnalysis.md
```diff
@@ -263,8 +263,53 @@ It is not already clear which new or updated mechanisms/components would mitigat
...
+
+## Requirements
+
+### Functional
+
+- *REQ-BlsKeyGenSecure*.
+Provide secure key generation with strong randomness requirements, resistance to side-channel leakage.
+- *REQ-BlsVariantAbstraction*.
+Support both BLS variants—small public key and small signature—behind a single abstraction. Public APIs are variant-agnostic.
+- *REQ-BlsSkToPk*.
```
curiecrypt, line 280: This requirement could be handled in *REQ-BlsKeyGenSecure*. Semantically, key generation covers both `sk` generation and the `pk` derivation from `sk`.

docs/ImpactAnalysis.md
```diff
@@ -263,8 +263,53 @@ It is not already clear which new or updated mechanisms/components would mitigat
...
+- *REQ-BlsDeterminismPortability*.
+Deterministic results across platforms/architectures; outputs independent of CPU feature detection.
+
+### Remarks
+
+- Note that the PoP checks probably are done at the certificate level, and that the above-described API should not be responsible for this.
+- The current code on BLS12-381 already abstracts over both curves `G1`/`G2`, we should maintain this.
+- The `BLST` package also exposes fast verification over many messages and signatures + public keys by doing a combined pairing check. This might be helpful, though it's currently unclear if we can use this speedup. It might be the case, since we have *linear Leios, that this is never needed.
 
 # Appendix
```
curiecrypt, line 314: - We should explain how the functional requirements map onto the primitives of the voting mechanism. An appendix section that describes these requirements in more detail, with references to the relevant parts of the CIP and specifications, would be valuable. For instance, domain separation is mentioned in *REQ-BlsSignVerify*, but its role within the Leios voting mechanism is not fully clear. An appendix clarifying such points would help the reader. Keeping these explanations in the appendix, while providing pointers in the requirements section when needed, would keep the main text simple yet informative.
- Another appendix section pointing out the risks and the potential mitigation strategies could be added.
- An explanation of the data flow with the rest of the code could be added. For example, while we already have serde functionality for the primitives, the way this data is transferred may raise additional considerations. We also need to account for the use of external data (e.g., keys, block hashes, etc.), which could have significant impacts on both Cardano Base and the cryptography component. It may not be possible to fully evaluate these impacts at this stage, but at minimum we should highlight the considerations and note their potential relevance for future analysis.

docs/ImpactAnalysis.md
```diff
@@ -263,8 +263,53 @@ It is not already clear which new or updated mechanisms/components would mitigat
...
+Support both BLS variants—small public key and small signature—behind a single abstraction. Public APIs are variant-agnostic.
+- *REQ-BlsSkToPk*.
+Deterministic sk → pk derivation for the chosen variant.
+- *REQ-BlsSignVerify*.
+Signature generation and verification APIs, variant-agnostic and domain-separated (DST supplied by caller). Besides the DST, the interface should also implement a per message augmentation (as the hash to curve function also has in the IETF draft)
+- *REQ-BlsPoP*.
+Proof-of-Possession creation and verification to mitigate rogue-key attacks.
+- *REQ-BlsAggregateSignatures*.
+Aggregate a list of public keys and signatures into one
+- *REQ-BlsBatchVerify*.
```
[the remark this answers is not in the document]
perturbing: I think the node will have two contexts/actions in which it will perform aggregation/batching:

1. If the node is creating the certificates, it is aggregating the signatures of the set of persistent signers.
2. If the node is verifying the certificate, it is aggregating the public keys and signatures of the persistent voters in the bitmap (see this question) and the non-persistent voters. Note that they sign over the same message, so the fastest verification is just by curve addition rather than repeated pairing calls.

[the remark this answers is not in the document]
curiecrypt: As we discussed today, we may need a tag to distinguish between signatures generated by persistent and non-persistent voters. For example:
- Persistent voters: "PV || MSG"
- Non-persistent voters: "NV || MSG"

Even so, if we consider the fields of Leios certificate given in [here](https://github.com/cardano-scaling/CIPs/blob/leios/CIP-0164/README.md#:~:text=size%20in%20bytes%0A%20%20%5D-,Votes%20and%20Certificates,-leios_certificate%20%3D%0A%20%20%5B%20election_id%20%20%20%20%20%20%20%20%20%20%20%20%20%20%3A%20election_id) we will aggregate the signatures generated by these parties separately. In this case, we should consider having the batch verification, which I think is not the highest priority at this time.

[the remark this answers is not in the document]
perturbing: What is the reason for such separation?

perturbing:
Oops, I accidentally closed it ;)

Issue #546, opened 2025-09-10
Impact analysis document


ch1bo:
## Why

- Because we promised intersect
- It is useful to gauge the complexity of changes we expect throughout the reference implementation (the haskell node)
- Will informa a more detailed node-level technical design 

## What

- Rough cardano-node component level impact of the changes as proposed in the CIP
- Ecosystem impact of baseline leios (as proposed) vs. the sharding ones (as done by serokell)
- ...

## How

- A markdown document published on the ouroboros-leios repository

mgmeier:
Performance & Tracing impact analysis PR: #561

PR #560, opened 2025-09-29
Impact analysis onto ecosystem

ch1bo:
Part of #546


docs/ImpactAnalysis.md
```diff
@@ -28,12 +28,130 @@ This document is the first iteration of a high-level design document for the Lei
...
+
+## Impact on operations
+
+The Leios upgrade does not change the roles and responsibilities for SPOs and only small changes to their operations are expected. Besides a modest increase in compute and network requirements (see [resource requirements summary in the CIP](https://github.com/cardano-scaling/CIPs/blob/leios/CIP-0164/README.md#resource-requirements)), one additional step in the block producing node setup is required: One additional key pair is needed to sign votes for EBs.
+
+- **REQ-GenerateBLSKeys** SPOs must be able to generate BLS key pairs.
+- **REQ-RegisterBLSKeys** SPOs must be able to register their BLS public key to become part of the voting committee.
+- **REQ-RotateBLSKeys** SPOs must be able to rotate their BLS key and limit usage of compromised keys.
+
+Concretely, these BLS keys could be generated and managed alongside the existing VRF keys. The node configuration needs to be extended to include the BLS signing key and the public key will need to become part of the operational certificate. With the signing key available, the node will be able to automatically issue votes for EBs.
```
nfrisby, line 43: > public key will need to become part of the operational certificate.

The opcert is part of the Praos header. So does this line imply the Praos header needs to include an entire public key? That'd be a non-trivial increase. And it's not listed here https://github.com/cardano-scaling/CIPs/blob/leios/CIP-0164/README.md#ranking-blocks-rbs

So maybe this is miswording?

docs/ImpactAnalysis.md
```diff
@@ -28,12 +28,130 @@ This document is the first iteration of a high-level design document for the Lei
...
+
+## Impact on user experience
+
+End-users are not expected to be impacted by the Leios upgrade. The transaction format and ledger semantics remain unchanged. Functionally, end-users will be able to use wallets and dApps as before, while dApp developers can continue to build on top of Cardano as before. However, throughput increases often come with a trade-off in latency. The proposed Leios upgrade only increases latency for transactions that need to be processed via EBs in times of higher than "Praos-only" load. The CIP provides more detail on expected latency under varying load [in the simulation results](https://github.com/cardano-scaling/CIPs/blob/leios/CIP-0164/README.md#simulation-results). Two key non-functional requirements for end-users are:
```
nfrisby, line 47: > The proposed Leios upgrade only increases latency for transactions that need to be processed via EBs in times of higher than "Praos-only" load.

I've heard this a couple times now, and it doesn't make any sense to me that Leios could be considered to have increased the latency compared to Praos.

Suppose that some subset of users wants to submit 12 megabytes of txs right now. Today without Leios, that'll take around ~150 blocks to get onto the chain. So the last of those txs will have huge latency. With Leios, that could instead happen as soon as one EB makes it on chain; say 5-10 RBs later. It seems reasonable to me to consider this as Leios _reducing_ latency.

On the other hand, if the system is not at all congested---ie the RBs themselves can host all of the existing tx traffic---then latency is the exact same as in Praos.

---

What am I missing?

docs/ImpactAnalysis.md
```diff
@@ -28,12 +28,130 @@ This document is the first iteration of a high-level design document for the Lei
...
+
+Both client interfaces provide access to the chain as a sequence of blocks: N2C via the `LocalChainSync` (see [ogmios API reference](https://ogmios.dev/mini-protocols/local-chain-sync/)) and UTxO-RPC via the [Sync Module](https://utxorpc.org/sync/intro/). As the Leios upgrade will change the block format, those APIs will need to be updated to reflect the new block structure. However, the fundamental concept of a block containing transactions remains unchanged and a backwards compatible representation of blocks is possible. For example, the `LocalChainSync` protocol is serving currently Praos blocks, whose body is basically a list of transactions (see the [Conway CDDL](https://github.com/IntersectMBO/cardano-ledger/blob/master/eras/conway/impl/cddl-files/conway.cddl)). With Leios, this would change to either a list of transactions or a certificate. However, a backwards compatible way would be to "inline" the transactions referenced via an EB in the block body, so that clients can continue to consume transactions directly from the block body. Two things need to be considered with this approach:
```
nfrisby, line 69: > a backwards compatible representation of blocks is possible

Is there idea here that the N2C ChainSync mini-protocol could simply serve a CertRB by inlining the certified EB into it as if it were a _huge_ TxRB?

I could imagine how that would be useful in some scenarios, but I also expect it would cause some (perhaps acceptable) confusion. Such a faked-up mega-RB will at least violate the per-RB limits on tx measurement dimensions (bytes, ExUnits, refscript bytes) and won't have the size nor the hash claimed by its Praos header.

An alternative is for N2C ChainSync so simply serve `(Maybe EB, RB)`. We could provide the `fakeUpMegaRB :: (Maybe EB, RB) -> RB` function (in a few popular languages) for developers to use as short-term integration hack, maybe?

nfrisby: Ah, I see now that I jumped the gun writing this comment :D

docs/ImpactAnalysis.md
```diff
@@ -28,12 +28,130 @@ This document is the first iteration of a high-level design document for the Lei
...
+
+1. The block body would grow significantly in size, as it would contain all transactions referenced by the EB. Client applications need to be handle this 10x or more increase in block size.
+
+2. The block body hash in the header would not be a simple hash of the body bytes anymore. However if the certificate is also part of the body, block integrity can be verified, albeit requiring a slightly more complex procedure.
+
+In summary the following requirements arise for client interfaces:
+
+- **REQ-ClientInterfaceChainSync** Client interfaces for chain synchronization and accessing block data should be updated in a backwards compatible.
```
nfrisby, line 77:
```suggestion
- **REQ-ClientInterfaceChainSync** Client interfaces for chain synchronization and accessing block data should be updated in a backwards compatible way.
```

docs/ImpactAnalysis.md
```diff
@@ -28,12 +28,130 @@ This document is the first iteration of a high-level design document for the Lei
...
+As the CIP points out, the Leios upgrade may increase transaction inclusion latency in times of high load. Some systems might be sensitive to this higher latency. This includes systems that need to perform conversions between assets on different chains or systems that need to react to on-chain events within a certain time frame. Examples of such systems are:
+
+* Bridges
+* DEXes
+* Algorithmic stablecoins
+* Layer 2 solutions
+
+#### Higher throughput
+
+Lastly, the mere increase in throughput may impact some systems. Currently, the Cardano network maximum data throughput capacity is 4.5 TxkB/s. The Leios protocol upgrade and proposed parameters may result in up to 250 TxkB/s - corresponding to a more than 50x increase. Besides transactional throughput, this also results in higher storage requirements for systems that index and provide access to chain data. While some system architecturs may have been designed with an order of magnitude more throughput in mind, 50x would be significant enough to result in unexpected load and may require optimizations or architecture changes. A non-functional requirement to mitigate this risk is:
```
nfrisby, line 145:
```suggestion
Lastly, the mere increase in throughput may impact some systems. Currently, the Cardano network maximum data throughput capacity is 4.5 TxkB/s. The Leios protocol upgrade and proposed parameters may result in up to 250 TxkB/s - corresponding to a more than 50x increase. Besides transactional throughput, this also results in higher storage requirements for systems that index and provide access to chain data. While some system architectures may have been designed with an order of magnitude more throughput in mind, 50x would be significant enough to result in unexpected load and may require optimizations or architecture changes. A non-functional requirement to mitigate this risk is:
```

nfrisby (approved):
Two comments that might provide food for thought and also a couple typos.

It's very nice to see an overview of the downstream impacts 👏👏👏

docs/ImpactAnalysis.md
ch1bo: Good catch. As you can tell, I'm not (yet) very familiar with the opcert situation. Having the BLS public key in each block header is probably not needed.. so where should it go, how should registration be achieved? Maybe we don't know and I'll just drop the sentence?

docs/ImpactAnalysis.md
ch1bo: You are right! I was using a too narrow view on latency per transaction here, considering the latency from (front of the) mempool to ranking block, which would be higher if we need to go the Leios route of announce & certify. However, if one considers the mempool as a whole, incorporating overall demand as your explanation does, the systemic latency will be lower .. even when taking the end-user's perspective! I'll adapt this paragraph with your input, thanks.

docs/ImpactAnalysis.md
ch1bo: Yeah.. exactly below there is this concern :D At this point we should be fine with pointing this out and see how client interfaces fall into place (how would the ideal Leios client API look like?).

docs/ImpactAnalysis.md
ch1bo: I tried to re-frame our analysis on this in https://github.com/input-output-hk/ouroboros-leios/commit/b8afe799fe52b5e75c84d26efbd97b791f0850ad

PR #563, opened 2025-10-01
End-to-end testing impact analysis

mkoura:


ch1bo (approved):
Thanks for contributing this, I'll merge it and might update it slightly to make it consistent with other chapters later.

PR #577, opened 2025-10-12
Impact analysis: Edit consensus, network, cryptography and other sections

ch1bo:
This makes language and form more consistent by grouping requirements in topic-based sections and keeping focus on what requirements emerge, only alluding to changes/risks and experiments where it makes sense.

Note: I used an LLM to consolidate sections giving instructions about form and language. I did a quick on correctness already and will do a self-review with the PR open too.


nfrisby (approved):
LGTM

I agree with the benefits of matching the shape of other content.

PR #578, opened 2025-10-13
Site/community

will-break-it:


ch1bo (changes requested):
We should rather start 2-3 threads on the CIP itself or a discussion in [this repo](https://github.com/input-output-hk/ouroboros-leios/discussions) or on the [cardano-foundation/CIPs](https://github.com/cardano-foundation/CIPs/discussions)

Both of which would be open forums to have a scientific argument and discussion. Any outcomes can then go into the CIP or onto our website (e.g. update in the FAQ section).

Issue #542, opened 2025-09-09
Write technical design and implementation plan


ch1bo:
## Why

The CIP is a protocol-level specification of _what_ the Leios protocol could be (and _why_ it is a good idea). It only briefly touches on what the concrete changes on a Cardano node would be. While we want to also produce implementation-independent documentation, a node-level look at _how_ Leios could be implemented is a key artifact to motivate and guide the next steps. Also, it will serve as a deliverable to Intersect.

## What

Topics to cover:
- Summarize the CIP and its impact onto the node (further refined from #546)
  - Concrete changes to node architecture, supported by a C4 component diagram?
- Identify/list key risks or assumptions to validate early
- Motivate a rough simulations -> prototype -> testnet plan
- What kind of prototype should we build?
- What experiments / tests should we conduct (on a testnet)?
- Synergies with Peras
- Interactions with Genesis (catching up node)
- Key generation, registration, rotation (in more detail)
- ...

## How

- A markdown document on the leios github repository should be sufficient
- ...

Tasks

- [ ] Render it into a PDF to follow suite with peras: https://tweag.github.io/cardano-peras/peras-design.pdf
- [x] #581
- [x] Create a C4 component diagram with the component changes
- [ ] Contribute technical details about networking @coot 
- [ ] Contribute technical details about votes & certificates @curiecrypt
- [ ] Contribute "what role formal methods play" in the implementation plan @ramsay-t 
- [ ] Write implementation plan section @ch1bo 
- [ ] Write interactions with other protocols sections @ch1bo 
- [ ] Write risks & mitigations section @ch1bo 
- [ ] Write introduction/overview section @ch1bo 
- [ ] ...

ch1bo:
@nfrisby you started to think a lot into this direction already when doing the leios portions on #546, could you help me characterize the _what_ a technical design document could contain? I was thinking that key threats to mitigate and how that motivates the implementation plan could go here?

nfrisby:
Apologies for the slow response; I've started the month very focused on the October demo.

- I do see a lot of overlap with `ImpactAnalysis.md`.
- However, `ImpactAnalysis.md` is largely a bag of requirements/changes/etc.
- I would like `TechnicalDesign.md` to perhaps share that bag of requirements but then to do a better job motivating the changes.
- And that would especially include introducing useful concepts (incl. examples), etc.
[...]

PR #599, opened 2025-10-28
Leios design: Cryptography - Voting, Certificate

curiecrypt:
A draft of the technical design for cryptographic primitives for the voting mechanism and certificate scheme. 

This PR contributes to #542.

### DIFF
--- a/docs/leios-design/README.md
+++ b/docs/leios-design/README.md
@@ -814,7 +814,25 @@ The requirements are organized into two main categories: **core functionality**,
 
 - **Proof-of-Possession** 
   - The `blst` C library does not provide a direct interface for generating a proof of possession ($PoP$). 
-  - This functionality must be implemented manually in `cardano-base`, leveraging the existing `blst` bindings to construct a $PoP$ from the available primitives.
+  - This functionality must be implemented manually in `cardano-base`, leveraging the existing `blst` bindings and will be exposed as part of the BLS signing interface.
+  - For each BLS keypair `(sk_bls, pk_bls)`, let `pk_bytes` be the canonical compressed encoding of `pk_bls`.  
+  - The PoP is defined as a BLS signature over the fixed message: `msg_pop = “PoP” || pk_bytes` where `"PoP"` is a fixed ASCII tag. This ensures that PoP messages are clearly separated from vote messages and cannot be confused with them.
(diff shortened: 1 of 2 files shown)


docs/leios-design/README.md
```diff
@@ -616,46 +616,227 @@ Corresponding types with instances of `EncCBOR` and `DecCBOR` must be provided i
...
+- Non-persistent votes include the Pool ID and an eligibility signature on the election ID.
+
+#### Certificate Generation
+- Upon receiving votes, both voter identities and their signatures are verified.  
+- Non-persistent votes are further validated for eligibility.  
+- Once a quorum of valid votes is collected, a certificate can be generated.  
+- The certificate includes:  
+  - The election ID and the message (the hash of the EB)  
+  - The identities of participating voters  
+  - Eligibility proofs for non-persistent voters
```
hjeljeli32: I assume that eligibility proofs are stored in the certificate as they are, they are not aggregated, right?

docs/leios-design/README.md
```diff
@@ -616,46 +616,227 @@ Corresponding types with instances of `EncCBOR` and `DecCBOR` must be provided i
...
+- **Public key derivation** 
+  - Implement deterministic derivation of the public key ($pk$) from the corresponding secret key ($sk$) for the selected BLS variant. 
+  - This requires adding a FFI binding to the `blst` library to enable secure and efficient key derivation within `cardano-base`.
+    ```C
+    void blst_sk_to_pk_in_g1(blst_p1 *out_pk, const blst_scalar *SK);
+    void blst_sk_to_pk_in_g2(blst_p2 *out_pk, const blst_scalar *SK);
+    ```
+
+- **Signature** 
+  - Implement signature generation and verification APIs that are variant-agnostic and support domain separation, with DST supplied by the caller. This functionality requires integrating with the `blst` library through FFI bindings, using the following functions:
```
hjeljeli32: It is great that you covered the DST. Do you think we should also cover the AUG (Augmented BLS signature scheme). I understood that it binds the signature to the signer’s public key — preventing certain rogue-key attacks in multi-signature settings (e.g., when aggregating signatures from multiple participants).

hjeljeli32 (approved):

perturbing:
@ch1bo , will do. Given our recent discussion, I think we can make some statements more exact and precise

PR #635, opened 2025-12-01
Leios design: document choice of BLS minSIG variant

hjeljeli32:
Follow-up to #621 on top of curiecrypt/leios-design-votes-certs.

This PR:
- Adds a “Choice of BLS variants (minPK vs minSIG)” subsection under Design choices.
- States explicit key/signature sizes for minPK and minSIG.
- Explains why minSIG is preferred, given Leios processes far more signatures than public keys.

### DIFF
--- a/docs/leios-design/README.md
+++ b/docs/leios-design/README.md
@@ -707,49 +707,68 @@ Beyond Leios, the BLS mechanism is also relevant to other Cardano subsystems; Mi
 #### Choice of BLS Variant
 
 BLS12-381 signatures can be instantiated in two variants that differ only in which group is used for public keys and which is used for signatures. Both variants are equivalent in security and share the same API surface; they differ only in the size of the encoded artifacts:
-- **MinPk variant**: public keys are 48 bytes, signatures are 96 bytes, and a proof of possesion is 2 times 96 bytes.
-- **MinSig variant**: signatures are 48 bytes, public keys are 96 bytes, and a proof of possesion is 2 times 48 bytes.
+- **MinPk variant**: public keys are 48 bytes, signatures are 96 bytes, and a proof of possession is 2 times 96 bytes.
+- **MinSig variant**: signatures are 48 bytes, public keys are 96 bytes, and a proof of possession is 2 times 48 bytes.
 This creates a straightforward trade-off: either public keys are smaller (MinPk) or signatures are smaller (MinSig).
 
-For Leios, this choice affects the size of voting certificates, which in turn impacts bandwidth, storage, and block propagation. Note that proofs of possession are only used at key registration time, so their size has negligible impact on steady-state bandwidth, storage, or verification costs. They therefore do not weigh into the choice between the two variants.
+For Leios, this choice affects the size of voting certificates, which in turn impacts bandwidth, storage, and block propagation. Note that proofs of possession are only used at key registration time, so their size has negligible impact on steady-state bandwidth, storage. They therefore do not weigh into the choice between the two variants.
 
-The certificate scheme we use follows the one defined in CIP-0164 together with the more detailed description in
-[Specification.md](https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/Specification.md) (section *“BLS certificate scheme”*). At a high level, a certificate records:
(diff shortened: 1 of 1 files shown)


docs/leios-design/README.md
```diff
@@ -702,6 +704,19 @@ The BLS voting mechanism relies on a pairing-based signature scheme defined over
 This approach is advantageous because it enables the aggregation of public keys and signatures, allowing large groups of participants to collectively signal approval through a single compact artifact. 
 Beyond Leios, the BLS mechanism is also relevant to other Cardano subsystems; Mithril already employs BLS-based aggregation, and Peras is expected to adopt a similar approach in future implementations.
 
+#### Choice of BLS Variant
+
+BLS12-381 signatures can be instantiated in two variants that differ only in which group is used for public keys and which is used for signatures. Both variants are equivalent in security and share the same API surface; they differ only in the size of the encoded artifacts:
+- **MinPk variant**: public keys are 48 bytes, signatures are 96 bytes.
+- **MinSig variant**: signatures are 48 bytes, public keys are 96 bytes.
```
perturbing:
```suggestion
- **MinSig variant**: signatures are 48 bytes, public keys are 96 bytes , and a proof of possesion is 2 times 48 bytes.
```

docs/leios-design/README.md
```diff
@@ -702,6 +704,19 @@ The BLS voting mechanism relies on a pairing-based signature scheme defined over
...
+This creates a straightforward trade-off: either public keys are smaller (MinPk) or signatures are smaller (MinSig).
```
perturbing: Perhaps add a note that the impact of having a bigger or smaller PoP is not weighing in the decision, because these are only used once at the registration level. But the impact of using min PK on the registration certificates is now clearer.

perturbing (approved):
Small comments on the PoP and its sizes :)

hjeljeli32:
> Thanks @hjeljeli32. I concur with this analysis. If the others agree, let's do:
> 
> * update the CIP, staying high-level, with definite specifications for vote and certificate structures (CDDL, amend the specification section as needed)
> * incorporate the analysis and any context from the CIP and https://github.com/input-output-hk/ouroboros-leios/blob/main/crypto-benchmarks.rs/Specification.md#bls-certificate-scheme into the `leios-design`, such that: it is not repeating what is already specified in the CIP, but extending it. The idea is that one only needs to have the CIP and `leios-design` to implement it correctly for the `cardano-node`. This means that we might want to concretize the general description of what vote / certificate contains with a Haskell data type definition, or at least enumerate the `cardano-base` types and methods that shall be used for signatures as defined in the CIP (e.g. the CDDL)

Hi @ch1bo ,
I’ve now updated CIP-0164 and also extended the Leios design README to integrate the latest analysis and quantify the MinSig vs MinPk trade-off.
Next, I’ll open a small PR to update Specification.md in the crypto-benchmarks so that the spec, the CIP, and the design document all remain aligned.
Thanks a lot for your suggestions!

docs/leios-design/README.md
```diff
@@ -709,13 +709,59 @@ Beyond Leios, the BLS mechanism is also relevant to other Cardano subsystems; Mi
...
+
+  \[ certificate_bytes = 40 + 2 * 48 + 113 + (28 + 48) * 100 = 136 + 113 + 7600 = 7849 bytes.
+  \]
+
+- **MinPk**:
+
+  \[ certificate_bytes = 40 + 2 * 96 + 113 + (28 + 96) * 100 = 232 + 113 + 12400 = 12745 bytes.
+  \]
+
+So, for n = 1000 and \(n - m ≈ 100\), MinPk certificates are about 4.9 kB larger, which is roughly a **60% increase** over MinSig.
```
ch1bo: The argument _could_ not only be about certificate size, but also about the size of votes sent around the network. In the end it's a no-brainer to go with `MinSig`, but maybe worthwhile to also point out the amount of traffic per round this saves.

hjeljeli32: @ch1bo 
Thanks a lot for the suggestion about vote traffic, that is a very good point.
I added an “Impact on vote size” subsection that quantifies the difference between MinSig and MinPk for both persistent and non-persistent votes.
This makes the bandwidth argument for MinSig more explicit.

PR #648, opened 2025-12-04
Specify MinSig variant and clarify non-aggregation of eligibility proofs in specification.md

hjeljeli32:
This PR updates crypto-benchmarks.rs/Specification.md to align it with:
- CIP-0164 (wire format + certificate structure)
- The updated Leios design README (BLS MinSig rationale)

Summary of changes
1. Explicitly state that Leios uses the BLS12-381 MinSig variant.
This matches the performance rationale (smaller signature size, significantly smaller certificates)
2. Clarify that eligibility proofs are not aggregated in Leios.
     - Each non-persistent voter produces an individual BLS signature (48B).
     - These proofs must be verified independently to enforce the sortition threshold.

@bwbush I’d appreciate your confirmation that:
- specifying MinSig explicitly here is correct and consistent with the intended benchmarks, and
- the clarification that eligibility proofs remain non-aggregated matches your understanding of the current design.

Thank you!

### DIFF
--- a/docs/leios-design/README.md
+++ b/docs/leios-design/README.md
@@ -763,7 +763,7 @@ We also evaluated the 500-seat case. The percentage increase from MinSig to MinP
 
 In summary, the Leios cryptography design therefore **fixes MinSig as the BLS12-381 variant**.
 
-### Implementation Plan
+### Implementation
 
 To implement the linear Leios design, efficient BLS signature functionality is essential for achieving fast and compact certificate generation. 
 With the adoption of [CIP-0381](https://cips.cardano.org/cip/CIP-0381), `cardano-base` already provides foundational utilities for BLS operations, offering a solid basis for this integration. 


docs/leios-design/README.md
```diff
@@ -616,46 +618,316 @@ Corresponding types with instances of `EncCBOR` and `DecCBOR` must be provided i
...
+
+**Interpretation for Leios**
+
+- With MinSig, certificates stay comfortably below **8 kB** and keeps certificate propagation and storage lightweight.
+- With MinPk, certificates grow to around **12.7 kB**, still within Praos limits but significantly heavier for bandwidth, storage, and diffusion—especially given the number of certificates the system will handle over time.
+- Since Leios produces far more **signatures** (votes and eligibility proofs) than public keys, the overall cost is driven by signature size rather than key size.
+
+In summary, the Leios cryptography design therefore **fixes MinSig as the BLS12-381 variant**.
+
+### Implementation Plan
```
ch1bo:
```suggestion
### Implementation
```

It's not really a plan but how the concrete functionality is to be implemented in `cardano-base` (as the intro of the section says nicely)

PR #682, opened 2025-12-09
Leios demo with configurable network delay and rate 

bladyjoker:
DONE

- [x] Configure `ip netns`
- [x] Configure `tc`
- [x] Update process-compose config
- [x] Update x-ray integration

### DIFF
--- a/demo/2025-11/README.md
+++ b/demo/2025-11/README.md
@@ -49,7 +49,6 @@ Run the Leios X-Ray (Grafana based observability stack)
 
 ```shell
 export LOG_PATH=".tmp-leios-202511-demo/*.log"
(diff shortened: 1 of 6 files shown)


demo/2025-11/scripts/init.sh
```diff
@@ -0,0 +1,55 @@
...
+  DELAY_N0_TO_UP
+  RATE_N0_TO_DOWN
+  DELAY_N0_TO_DOWN
+  RATE_DOWN_TO_N0
+  DELAY_DOWN_TO_N0
+)
+
+for var in "${VARS_TO_DEFAULT[@]}"; do
+  def_var="DEF_$var"
+  val=${!var:-${!def_var}}
```
ch1bo: This is a bit odd, why don't we have defaults in the various scripts and just do overrides via `.env`? or why are there even `DEF_XXX` for `XXX` and not just set the vars to defaults .. by default :sweat_smile: ?

PR #751, opened 2026-01-29
Site: remove weekly updates

ch1bo:
Those are not updated anymore. I have redirected the URL to the monthly-reviews, cleaned up the navbar and re-ordered entries in the sidebars


berewt (approved):
I think we're good. I don't think there any reason to keep the weekly notes in main, the git history is indeed enough here.

PR #798, opened 2026-03-05
Fix network emulation

crocodile-dentist:
Reworked setting up the network topography

### DIFF
--- a/demo/burst/scripts/init-namespaces.sh
+++ b/demo/burst/scripts/init-namespaces.sh
@@ -1,115 +1,82 @@
 set -exuo pipefail
 
+IP=10.0.0.
+#upstream is 10.0.0.1
+#middle/N0   is 10.0.0.2
+#downstream is 10.0.0.3
+
+DELAYS1=([2]=$DELAY_UP_TO_N0)
+RATES1=([2]=$RATE_UP_TO_N0)
+DELAYS2=([1]=$DELAY_N0_TO_UP [3]=$DELAY_N0_TO_DOWN)
+RATES2=([1]=$RATE_N0_TO_UP [3]=$RATE_N0_TO_DOWN)
+DELAYS3=([2]=$DELAY_DOWN_TO_N0)
+RATES3=([2]=$RATE_DOWN_TO_N0)
+
+HOSTS=$(seq 1 3)
+
 NS_PREFIX="leios-experiment"
-NS_UPSTREAM="${NS_PREFIX}:upstream"
-NS_NODE0="${NS_PREFIX}:node0"
-NS_DOWNSTREAM="${NS_PREFIX}:downstream"
+NS=([1]="${NS_PREFIX}:upstream" "${NS_PREFIX}:node0" "${NS_PREFIX}:downstream")
 
-# Delete all namespaces
-ip netns del "$NS_UPSTREAM" || true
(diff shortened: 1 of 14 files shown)


demo/burst/scripts/init-namespaces.sh
```diff
@@ -55,41 +53,21 @@ for i in $HOSTS; do
...
-done
-for src in $HOSTS; do
-    declare -n DELAYS="DELAYS$src"
-    for dst in ${!DELAYS[@]}; do
-        set -x
-        DELAY=${DELAYS[$dst]}
-        SRC_IP=10.0.0.$src
-        tc -n ns-br filter add dev veth${dst}-br protocol ip parent 1: prio 1 u32 \
-           match ip src $SRC_IP flowid 1:$src
-        tc -n ns-br qdisc add dev veth${dst}-br parent 1:$src netem delay ${DELAY}ms
```
crocodile-dentist: There's a technical reason why it's a good idea to attach `netem` to `prio` instead of directly to the bridge. This part should be reverted.

demo/burst/scripts/init-namespaces.sh
```diff
@@ -55,28 +55,41 @@ for i in $HOSTS; do
...
     set -x
     tc -n $NS_ qdisc add dev veth$i root handle 1: htb
-    tc -n ns-br qdisc add dev veth${i}-br root handle 1: prio
     { set +x; } 2>/dev/null
     for peer_x in ${!DELAYS[@]}; do
         set -x
         IP=10.0.0.$peer_x
-        DELAY=${DELAYS[$peer_x]}
         RATE=${RATES[$peer_x]}
-        JITTER=$(( DELAY * 15 / 100 ))ms
```
crocodile-dentist: Let's leave the JITTER in, it's not used now but it is important for realistic modeling.

crocodile-dentist:
Commit [demo: fix network delay not being applied in bridge-based TC setup](https://github.com/input-output-hk/ouroboros-leios/pull/798/commits/6462a3b672182fb82633cc8c2d0fe5a7db18f6ff) has this in the message:
>The previous approach attached netem qdiscs to the bridge-side veth
interfaces (veth${i}-br) using filters matching `ip dst PEER_IP`. This
never worked: egress traffic on veth${i}-br goes *toward* node i, so it
carries dst=IP_NODE_i, not dst=PEER_IP. The filters never matched, and
delay was only applied accidentally via the prio qdisc's default band
fallback — and only for nodes whose peer indices happened to land on that
default band. In a 3-node topology, node 2 (the middle node, with peers
1 and 3) consistently received no delay at all because neither peer index
mapped to the default priomap band (1:2).

But that is not true.

demo/burst/scripts/init-namespaces.sh
ch1bo: Addressed your concerns in https://github.com/input-output-hk/ouroboros-leios/commit/3bb642080231e8a314f4dc55849b37232ca0dce7

ch1bo:
@crocodile-dentist Consolidated only the changes you agreed with into https://github.com/input-output-hk/ouroboros-leios/commit/3bb642080231e8a314f4dc55849b37232ca0dce7. Hope this now not degrades the network emulation setup and just apply the fixes you did for `demo/burst` also to `demo/protot-devnet`.

Issue #805, opened 2026-03-05
Simulate different voting systems


sandtreader:
### Why

https://github.com/cardano-scaling/leios-wfa-ls-demo/issues/2 will give us numbers for the costs - CPU, size - of various persistent and non-persistent votes, but we need to simulate the traffic caused by different design decisions, particularly around committee size given the O(N^2) traffic.

There are three (arguably 2.5) options currently being considered:

A) wFA^LS with persistent (PV) and non-persistent (NPV) voters as described in the CIP
B) Everyone (absolutely everyone) votes (C=~3000, all PV)
C) A much smaller subset of everyone votes (C=~500, all PV)

The current Rust simulator (sim-rs) only handles votes in an abstract way and has no concept of different types.

### What

Add the concept of persistent and non-persistent votes to the simulator, parameterised by benchmark figures from https://github.com/cardano-scaling/leios-wfa-ls-demo/issues/2 and run simulations of the options.

### How

Migrate the simulation vote handling to multiple types
Create simulation configurations for the various options
Run and report on the outcomes

sandtreader:
Headline results for a 1500-node sim, running "everyone votes" (literally everyone, even relays), as reported in the April monthly:

Mode | Committee size | Total votes | Avg votes/EB | EBs meeting threshold | EBs endorsed | TX→EB (s)
-- | -- | -- | -- | -- | -- | --
wfa-ls | 600 expected (PV 480 + NPV 120) | 31,790 | 447.7 (σ 212.9) | 56/74 = 75.7% | 31/74 = 41.9% | 216.9 (σ 109.7)
everyone | 1500 (all nodes incl. 1042 relays) | 82,806 | 1119.0 (σ 512.7) | 56/74 = 75.7% | 31/74 = 41.9% | 217.0 (σ 109.8)

The key take-home is that even having 2.5x more voters has no effect on Leios liveness whatever, and hence (within reason) the choice of committee size is not a factor in choosing a committee selection algorithm.

Full results are documented in https://github.com/input-output-hk/ouroboros-leios/pull/882

PR #806, opened 2026-03-05
Split vote config into separate persistent and non-persistent voter parameters

sandtreader:
Replace pre-computed weighted averages (80:20 persistent:non-persistent) with independently configurable persistent and non-persistent voter parameters for vote generation CPU time, vote validation CPU time, and vote bundle size per EB:

```yaml
persistent-vote-generation-probability: 400.0
non-persistent-vote-generation-probability: 100.0

persistent-vote-generation-cpu-time-ms: 135e-3
non-persistent-vote-generation-cpu-time-ms: 280e-3

persistent-vote-validation-cpu-time-ms: 670e-3
non-persistent-vote-validation-cpu-time-ms: 1.4

persistent-vote-bundle-size-bytes-per-eb: 90
non-persistent-vote-bundle-size-bytes-per-eb: 164
```

The code now computes the weighted averages from these component values.

Old config fields are grandfathered for Haskell sim compatibility.

Part of #805

### DIFF
--- a/data/simulation/config.schema.json
+++ b/data/simulation/config.schema.json
@@ -528,7 +528,14 @@
     },
     "vote-bundle-size-bytes-per-eb": {
       "additionalProperties": false,
+      "description": "Grandfathered for Haskell sim.",
       "properties": {},
+      "type": "number"
+    },
+    "persistent-vote-bundle-size-bytes-per-eb": {
+      "type": "number"
+    },
+    "non-persistent-vote-bundle-size-bytes-per-eb": {
       "type": "number"
     },
     "vote-diffusion-max-bodies-to-request": {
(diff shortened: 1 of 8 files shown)


data/simulation/config.schema.json
```diff
@@ -528,9 +528,16 @@
     },
     "vote-bundle-size-bytes-per-eb": {
       "additionalProperties": false,
+      "description": "Grandfathered for Haskell sim.",
```
ch1bo, line 531: What does grandfathered mean?

ch1bo (approved):

data/simulation/config.schema.json
sandtreader: Backwards compatibility, basically.

ch1bo: That is a term i would have understood :)

PR #835, opened 2026-04-01
chore: Update antithesis Dockerfile git refs

wolf31o2:
This updates to the latest refs on the reference branches


ch1bo:
Is there a way we could parse the refs from the `flake.lock` and provide these as ARGs to the docker build?

wolf31o2:
Let me see what I can do.

wolf31o2:
Apparently, the flake.lock isn't as updated as the refs here are.

wolf31o2:
@ch1bo switched to tx-centrifuge and switched to `flake.lock` usage for versioning

ch1bo:
> Apparently, the flake.lock isn't as updated as the refs here are.

Sometimes that is desirable

PR #848, opened 2026-04-08
[DONOTMERGE] docs/leios-design/LeiosFetchMarch2026Checkpoint.md

nfrisby:
I'm using this PR to foster discussion of design ideas with the Networking Team.

### DIFF
--- a/docs/leios-design/LeiosFetchMarch2026Checkpoint.md
+++ b/docs/leios-design/LeiosFetchMarch2026Checkpoint.md
@@ -43,6 +43,14 @@ It's useful for this single document to collate their essential descriptions sin
     - It's not obvious what the tolerable performance lower bounds are nor how they should vary across time and/or peers.
     - **Constraint**: LeiosFetch must quickly react to changes in peer behavior.
 
+- **Challenge**: there are other mini protocols running concurrently, and so far the node's architecture has kept them independent.
+  The `ouroboros-network` Mux Layer combines their network traffic fairly.
+  The GHC run-time system and the node's careful use of synchronization primitives combines their CPU usages fairly.
+  But with the introduction of Leios and its intended throughput, mini protocols are no longer equal.
+  Leios itself is strictly lower-urgency than Praos (e.g. ChainSync and BlockFetch).
+  And the increased overall throughput means TxSubmission load is potentially much more heavy than before (e.g. refilling Mempool after an EB flushes some of its content).
+    - **Constraint**: LeiosFetch decision logic _may_ need to coordinate with other mini protocols' decision logic (e.g. to balance/schedule network load).
+
 - **Challenge**: intercontinental peers are important.
   Despite intercontinental peers inherently exhibiting poor latency, they drastically improve the diffusion of EBs through the global Cardano network when well-utilized.
   Thus, There is no one-dimensional metric that would be sufficient for peer performance: a variety of peers is desired.
@@ -113,6 +121,8 @@ Ultimately, Praos is lightweight enough to be less challenging.
 As a result, a greater fraction of a node's peers will each be performant enough to diffuse a Praos block within the required latency.
 The difference is unsurpising, since Leios so drastically increases throughput.
 
+TODO give some examples of e.g. _hedge requests_ in BlockFetch and TxSubmission today already
+
 # Prior Art: The Tail At Scale
 
 "[The Tail At Scale](https://research.google/pubs/the-tail-at-scale/)" by Jeffrey Dean and Luiz André Barroso (2013) discusses a similar goal: improving worst-case latencies among a large network of machines.
@@ -162,6 +172,8 @@ TODO find people with greater expertise---I'm a comparatively sheltered Ouroboro
   Major difference: The tree structure is a crucial element, and the existing Cardano p2p logic is not obviously compatible with that; e.g. that article mentions ~2 hops instead of our ~6.
   Major difference: The Solana objective includes much higher sustained traffic, so hopefully Linear Leios doesn't require as much complexity.
   Major difference: They're separately compensating for UDP issues instead of accepting TCP's limitations.
+
+TODO further discussion on the high-level trade-offs of erasure codes, esp. what resource usage does it improve the efficiency of versus how much CPU etc. it would consume?
 
 # Terminology
 
--- a/docs/leios-design/LeiosFetchMarch2026Checkpoint.md
+++ b/docs/leios-design/LeiosFetchMarch2026Checkpoint.md
@@ -1,0 +1,232 @@
+# Introduction
+
+This report summarizes the high-level challenges that the LeiosFetch decision logic faces.
+It also proposes an outline of the high-level rules that would navigate those challenges.
+
+This particular document has two concrete purposes.
+
+- Snapshot the high-level design of the LeiosFetch decision logic as of the end of 2026 March.
+- Catalyze critique and contributions from teammates and community members interested in the architecture.
+
+# High-Level Challenges
+
+The following list provides a terse but somewhat self-contained description of each challenge.
+Future writing will elaborate on each and their interactions.
+It's useful for this single document to collate their essential descriptions since LeiosFetch lies at their intersection.
+
+- **Challenge**: the stakes are high, even for rare events (a.k.a. "in the tail").
+  Unanticipated latency of EB diffusion risks increasing the latency of Praos messages.
+  If a Praos block contains a certificate for an EB, a node cannot fully select that Praos block before it receives that EB.
+  Thus, the absence of an EB could increase Δ_Praos (a.k.a. Δ_RB), thereby degrading Praos security.
+  This is the most important risk for Leios, since Praos is the foundation of not only the Leios security argument but the Cardano security argument.
+    - **Constraint**: The worst-case cannot be ignored.
+
+- **Challenge**: multiple hops.
+  The EB diffusion in question starts at the EB producer and ends at "every" other honest node (analyses can disqualify honest nodes if they're latency outliers and the honest stake ratio is high enough to afford their loss).
+  Previous work has reported that the shortest path usually involves ≤ ~6 hops.
+  Some extra complexity can be added to pipelining those hops, but it won't be perfect.
+  LeiosFetch only manages the communication along one hop and the next (i.e. both sides of a single node), with the performance constraint tight enough that a composition of ~6 hops still has low enough end-to-end latency (≤ ~14 seconds for a healthy EB, accoding to CIP-0164).
+    - **Constraint**: Per-hop latency needs to be quite short, ~2 to ~3 seconds.
+
+- **Challenge**: peers are permissionless.
+  Peers are identifiable only by their current IP address and port.
+  Some peers can be sampled from the stake pools' registration data, but not all peers---even upstream peers---are acquired that way.
+  (TODO is that fact a hard constraint on the design?)
+    - **Constraint**: There are too many potential peer names to definitely keep a significant history of performance/quality profiles.
+
+- **Challenge**: unpredictable latency and bandwidth, per peer.
+  A peer's past performance does not necessarily correlate with its future performance.
+  Even for completely trustworthy peers, the particular traffic route is dynamic and not necessarily reliable, since Cardano operates over the public Internet.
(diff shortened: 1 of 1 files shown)


nfrisby:
Rendered https://github.com/input-output-hk/ouroboros-leios/blob/nfrisby/march2026-checkpoint/docs/leios-design/LeiosFetchMarch2026Checkpoint.md

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+    - **Constraint**: There are too many potential peer names to definitely keep a significant history of performance/quality profiles.
+
+- **Challenge**: unpredictable latency and bandwidth, per peer.
+  A peer's past performance does not necessarily correlate with its future performance.
+  Even for completely trustworthy peers, the particular traffic route is dynamic and not necessarily reliable, since Cardano operates over the public Internet.
+  Moreover, not at all peers are trustworthy, so even if the physical performance hasn't changed, any peer could suddenly choose to respond much more slowly or not at all.
+  Timeouts limit the affect of so-called _low-and-slow_ attacks, but two principal difficulties remain.
+    - Timeout false alarms are harmful to the network topology and hence also delay EB diffusion, so the thresholds cannot be arbitrarily tight.
+    - It's not obvious what the tolerable performance lower bounds are nor how they should vary across time and/or peers.
+    - **Constraint**: LeiosFetch must quickly react to changes in peer behavior.
```
coot, line 44: > A peer's past performance does not necessarily correlate with its future performance.
  Even for completely trustworthy peers, the particular traffic route is dynamic and not necessarily reliable, since Cardano operates over the public Internet.

That's why the $\Delta_Q$ model is appealing.
* we measure `G` using keep-alive messages
* we use a degenerate distribution (as a crude approximation) with exponential moving average when combining new measurements

We cannot predict the future (e.g. changes in global traffic routing, etc), but if a connection is regularly disturbed, then it will be reflected in our $\Delta_Q$ model.

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+  Leios itself is strictly lower-urgency than Praos (e.g. ChainSync and BlockFetch).
+  And the increased overall throughput means TxSubmission load is potentially much more heavy than before (e.g. refilling Mempool after an EB flushes some of its content).
+    - **Constraint**: LeiosFetch decision logic _may_ need to coordinate with other mini protocols' decision logic (e.g. to balance/schedule network load).
+
+- **Challenge**: intercontinental peers are important.
+  Despite intercontinental peers inherently exhibiting poor latency, they drastically improve the diffusion of EBs through the global Cardano network when well-utilized.
+  Thus, There is no one-dimensional metric that would be sufficient for peer performance: a variety of peers is desired.
+    - This makes timeouts even more complicated to enforce.
+    - An intercontinental peer is best utilized by saturating its high Bandwidth-Delay Product (BDP), which requires proportional memory.
+    - **Constraint**: LeiosFetch should treat peers heterogenously (e.g. different BDPs).
```
coot, line 59: For Praos we didn't have to make timeouts tight.  Is there a reason why they should be tight for Leios?

It might be challenging to know BDP without disturbing the node.  Can we assume/require minimum bandwidth (the delay we already measure using $\Delta_Q$ model)?

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+  The Leios node has a limited amount of CPU, RAM, and local networking resources.
+  A typical node's connectivity involves ~25 upstream peers.
+  LeiosFetch must prevent adversarial peers---who cannot be identified a priori---from consuming too much more than their fair share of those resources.
+    - **Constraint**: LeiosFetch cannot simply request every EB offered.
+
+- **Challenge**: scarce resources per downstream peer.
+  The Leios node has a limited amount of CPU, RAM, and local networking resources.
+  Popular nodes will serve > 100 downstream peers at once, so the server-side logic's must utilize resources frugally.
+  LeiosFetch must prevent adversarial peers---who cannot be identified a priori---from consuming too much more than their fair share of those resources.
+    - **Constraint**: LeiosFetch cannot simply serve every EB it offers.
```
coot, line 88: An honest peer won't be able to consume every EB we offer at once, so the we don't need to serve every EB we offer instantly.  If an honest client is struggling to process blocks it will request EBs in a slower way (there is some lag introduced by already pipelined requests, so pipelining depth should be chosen with care).  It seems we just need to avoid serving more than a node could handle (e.g. some byte size limit on number of bytes in-flight per connection might be enough).

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+There are two key differences.
+- Their network of concern is operated by the protocol designers.
+  That is qualitatively different scenario than the "unpredictable latency and bandwidth", "peers are permissionless", "multiple hops", and "unpredictable patterns for short-term Leios load" challenges, etc.
+- Moreover, the stakes are lower.
+  The cost of inflated tail latency within a fleet is proportional to its extent and duration; that could be costly, but not existential.
+  The cost of Leios accidentally delaying some Praos blocks is instead whether or not the Cardano security argument has an unmitigated gap.
+
+Even so, this prior art is an appropriate place to anchor the LeiosFetch implementation's design.
+It describes four ideas that could be leveraged by the LeiosFetch logic.
+- _Hedge requests_ are intentionally redundant requests, sent because---even among a centralized fleet---it's unknown which peer will be first to successfully reply.
```
karknu, line 138: Mention that praos does hedge requests for blocks.

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+  The EB diffusion in question starts at the EB producer and ends at "every" other honest node (analyses can disqualify honest nodes if they're latency outliers and the honest stake ratio is high enough to afford their loss).
+  Previous work has reported that the shortest path usually involves ≤ ~6 hops.
+  Some extra complexity can be added to pipelining those hops, but it won't be perfect.
+  LeiosFetch only manages the communication along one hop and the next (i.e. both sides of a single node), with the performance constraint tight enough that a composition of ~6 hops still has low enough end-to-end latency (≤ ~14 seconds for a healthy EB, accoding to CIP-0164).
+    - **Constraint**: Per-hop latency needs to be quite short, ~2 to ~3 seconds.
+
+- **Challenge**: peers are permissionless.
+  Peers are identifiable only by their current IP address and port.
+  Some peers can be sampled from the stake pools' registration data, but not all peers---even upstream peers---are acquired that way.
+  (TODO is that fact a hard constraint on the design?)
```
karknu, line 34: No, it isn't a hard constraint on the design.

The ability to use unregistered peers as upstream peers was a requirement for on the network layer. It enables nodes that would other wise have been leaf nodes to fully participate. These nodes help the network by propagating blocks and TXs.
Unregistered nodes also provides a layer of protection for SPOs. An attacker may target their publicly registered relays but their blocks can still propagate through their unregistered relays.

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+- **Challenge**: scarce resources per downstream peer.
+  The Leios node has a limited amount of CPU, RAM, and local networking resources.
+  Popular nodes will serve > 100 downstream peers at once, so the server-side logic's must utilize resources frugally.
+  LeiosFetch must prevent adversarial peers---who cannot be identified a priori---from consuming too much more than their fair share of those resources.
+    - **Constraint**: LeiosFetch cannot simply serve every EB it offers.
+
+- **Challenge**: EBs are large enough that their resource usage does not trivially scale.
+  CIP-0164 bounds the size of an EB at ≤ 12.512 MB; just 80 maximal EBs would already exceed 1 GB.
+  For example, if a popular node were to relay one maximal EB to every downstream peer, that alone would consume > 1.2512 GB of egress bandwidth.
+    - **Constraint**: LeiosFetch cannot simply request every EB from every peer that offers it.
```
coot, line 93: We might want to support multiple NICs, this will not solve the challenge but it should allow larger egress bandwidth in more advanced setups.

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+- _Tied requests_ are a mechanism so that not all of the M requests will be processed.
+  Whichever request is processed soon enough might occur so much faster that some of the other requests could be cancelled.
+  The original requests themselves cannot be unsent, but a "cancel-if-you-haven't-already-started" message might arrive in time to relieve some load from an overloaded peer.
+- A _shadow request_ is additional hedge request beyond the M requests.
+  It is sent to a peer that is currently on _probation_/being _snubbed_ because its past performance fell below acceptable tolerances.
+  If their performance has since increased and their response is now competitive, their probation can be lifted.
+
+Shadow requests seem the least likely mechanism for LeiosFetch to adopt, since Cardano nodes typically simply disconnect when a peer performs below tolerances (e.g. times out).
+Perhaps an intermediate probation stage could be useful, though.
```
coot, line 148: If any error was thrown (including a timeout), we're reconnecting after a fixed [`daRepromoteDelay`](https://ouroboros-network.cardano.intersectmbo.org/ouroboros-network/Ouroboros-Network-Diffusion-Policies.html#v:repromoteErrorDelay) (`10s`).  The peer might not end up in `hot` peers immediately - if it had a good score, it will be promoted in the next churn round.  We might want a quicker way to get that peer to the `hot` set  if it's score was descent, and slowly degrade it over time if it's current performance is not good enough.

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+
+Tied requests cannot be implemented as in the paper.
+There, the concrete suggestion was two-fold.
+- Upstream peers communicate with each other in order to cancel requests, since that way the cancellation can be sent as soon as the first response is _sent_ rather than _received_.
+- Clients should send hedge requests only after a delay of "two times the average network message delay (1ms or less in modern data-center networks)".
+  That way, 95% of the time, the first server to succeed can send a cancellation before the other servers receive their copy of the client request.
```
coot, line 154: We measure round trip with `KeepAlive` messages, which can be used for the hedge request delay.

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+The recommendation of the average ping obviously conflicts with at least the "unpredictable latency and bandwidth" and "intercontinental peers are important" challenges.
```
coot, line 155: Since all intercontinental peers will get this penalty, this will not influence `upstreamyness` nor `fetchyness` metrics used by `churn`.  But there's a conflict, peers have an incentive to send their data as soon as possible to get a better score and remain `hot` at the other end (such connections disseminate own blocks).

docs/leios-design/LeiosFetchMarch2026Checkpoint.md
```diff
@@ -0,0 +1,244 @@
...
+TODO Additional thoughts.
+
+- TODO perhaps servers should tolerate some degree of optimism from client's, so that MsgLeiosBlockTxs might be allowed to arrive before the MsgLeiosBlockOffer is sent?
+  That would hide the round-trip time between sending an offer and receiving the corresponding request.
+
+- Possible optimization: notice when an already-sent chunk request becomes redundant due to _subsequent_ TxCache insertions, since doing so reduces the EB's latency.
+    - Sending the cancellation as well in that case might avoid wasted upstream load.
+      That's also good, but secondary.
+
+- In practice, however, an EB will always be offered by one peer (likely/intended to be an intercontinental peer for intercontinental EBs) before any other peer offers it.
```
coot, line 237: This is true for intercontinental EBs, but once they reach a continent (or a particular data center), soon there will be multiple peers offering them simultaneously in that location if there are only handful intercontinental connections from that location.

ch1bo:
Superseded by https://github.com/input-output-hk/ouroboros-leios/pull/1005

PR #851, opened 2026-04-10
Add adaptive key corruption scenarios to threat model

ch1bo:
These have been discussed while we also looked at various committee selection schemes.


jesusdiazvico:
LGTM!

docs/threat-model.md
```diff
@@ -314,3 +315,26 @@ Excessive chain growth poses a different challenge where the success of Leios co
...
+
+This applies to both BLS voting keys and VRF keys. For BLS keys, the adversary corrupts honest parties one at a time, silently collecting their voting keys until enough are held to meet the certificate quorum threshold. At that point, the adversary can forge a certificate on invalid transactions and post it on-chain. This is particularly dangerous because Praos assumes that transactions on the main chain are valid, even against adaptive adversaries. If Leios is not adaptive-secure, this assumption breaks: an adaptive adversary could use forged certificates to get invalid transactions included on-chain, undermining the safety guarantees that Praos provides. For VRF keys, a similar accumulation strategy allows the adversary to gain disproportionate eligibility for EB creation and voting committee selection without any grinding — the corrupted keys simply grant their rightful eligibility to the adversary.
```
rrtoledo: > For VRF keys, a similar accumulation strategy allows the adversary to gain disproportionate eligibility for EB creation and voting committee selection without any grinding — the corrupted keys simply grant their rightful eligibility to the adversary.

Perhaps I don't understand well "to gain disproportionate eligibility for EB creation", but for the grinding part I think it might be a bit more complicated than that here.

Grinding-wise, the attack is about submitting a block or not. Knowing the VRF key gives you the ability to make a more informed choice on whether you should not publish a block now to gain more blocks/EBs later on. 
If an adversary got knowledge of a honest party key, the honest party can still publish its block. So knowing a honest party key does not help in doing a grinding attack.

What it helps though, is in knowing the slot leader and voting schedule, with the former helping the adversary knowing if a grinding attack is viable.

edit: Perhaps you are saying that knowing a VRF faciliate a grinding attack that can increase eligibility for EB creation?

rrtoledo (commented):
I am not sure about the VRF part otherwise LGTM

docs/threat-model.md
ch1bo: It's orthogonal to any grinding. Undetectedly (because not yet used) compromise of keys is what I understood that these attack vectors to be about. If that is done for VRF key, the attacker would accumulate more opportunities to create praos or leios blocks.

rrtoledo: You could use, without disclosing, the VRF key locally to compute someone else slot leader or vote eligibility schedule though. That can be useful for grinding (knowing when i have a majority of RBs on Praos) or DoSing (either EB certifying or RB publication) to lower Cardano's throughput.

perturbing (approved):

docs/threat-model.md
ch1bo: I see.. I have mistaken that the VRF only allows us to see the eligibility, and not the actual block production (both, for Praos or Leios blocks). I'm going to update/remove this sentence.

PR #858, opened 2026-04-14
fix: handle transient errors and reduce poll frequency in wait-for-test

wolf31o2:
The completion poll was crashing on transient DNS/connection failures from moog because the error output couldn't be parsed by jq under set -e. Now all phases catch query and parse failures and retry. Completion poll interval increased from 60s to 300s since the test runs for an hour.


wolf31o2:
@copilot apply changes based on the comments in [this thread](https://github.com/input-output-hk/ouroboros-leios/pull/858#pullrequestreview-4105736565)

ch1bo (approved):

PR #872, opened 2026-04-23
Update the cost-estimate to CIP-164 protocol

ch1bo:
The cost-estimates are severely outdated and use the very old measure of "input blocks per second", let alone counting all the IBs and even ignoring that we'd need to store some of them even though they don't get to be "executed" on the final ledger state (redundant/conflicting txs problem).

Anyhow .. I have updated the analysed protocol component to match Linear Leios (as proposed in CIP-164) and recomputed figures together with a :genie:. I did go over the numbers once, but more review is always appreciated and/or take these numbers with quite some :salt:

The main changes are:

- Costs updated to reflect CIP-164 Linear Leios as simulated: the previous model was built around Short Leios with IBs; this aligns with the actual CIP-164 protocol (mempool-based tx gossip, EBs referencing by hash) and uses mainnet-measured CPU benchmarks, matching the CIP simulation results.
- CPU and RAM costs are flat across the full throughput range: no IB body validation means CPU stays at 2 cores throughout; UTxO-HD assumed for both Praos and Leios (4 GB RAM), consistent with the planned deployment order.
- Economic break-even corrected: a pre-existing 10× arithmetic error in the TPS values was fixed, and the ADA price assumption updated to $0.25 (April 2026), giving more realistic required throughput figures.

---

This now also includes a complete overhaul of the `cost-estimator` interactive calculator to use the same models (CIP-164) and a full re-design:

image


docs/cost-estimate/01-compute-cpu.md
```diff
@@ -27,180 +26,187 @@ comes from block validation:
...
 
-   $$C_{\text{cert}} = \frac{N_{\text{cert}} \times T_{\text{cert}}}{S_{\text{length}}}$$
-   where:
-   - $N_{\text{cert}}$ = Number of certificates per stage (1)
-   - $T_{\text{cert}}$ = Certificate validation time (140ms)
-   - $S_{\text{length}}$ = Stage length in seconds (20s)
+   $$C_{\text{reapply}} = N_{\text{cert}} \times \left(0.3539 + 0.00002151 \times \frac{\text{TxkB/s} \times 1{,}000}{N_{\text{cert}}}\right)$$
 
-5. **Total CPU Usage**:
+   Simplifying (where $N_{\text{cert}} = 0.024$ cert/s):
```
ch1bo: I think this is wrong: we are using means everywhere here (a block every 20s = 0.05 / s) and the expected value of certification with the expected value of block production (every 20s) is 1

docs/cost-estimate/02-compute-ram.md
```diff
@@ -1,185 +1,155 @@
...
-Leios uses key-value store technology with on-disk persistent storage for the majority of the ledger state.
-The implementation favors write-optimized structures (e.g., LSM trees) and keeps only critical components in memory.
+| TxkB/s      | Tx Cache Size               |
+|-------------|-----------------------------|
+| 4.5 (Praos) | — (not applicable to Praos) |
+| 5           | 18 MB                       |
+| 50          | 180 MB                      |
+| 100         | 360 MB                      |
+| 200         | 720 MB                      |
+| 300         | 1,080 MB                    |
```
ch1bo, line 71: This is probably too little as we recently learned, right @nfrisby ?

docs/cost-estimate/04-egress.md
```diff
@@ -1,412 +1,261 @@
...
+> → expected peers served = 20 × 10% = **2**. Same structure as block bodies
+> (downstream), just mirrored: 100 inbound × (1/20) = 5 vs 20 outbound × (2/20) = 2.
+>
+> The **2-peer fetch** reflects TxSubmission V2 (`txInflightMultiplicity = 2` in
+> `TxDecisionPolicy`), which deliberately downloads each tx from 2 upstream peers
+> simultaneously for redundancy. TxSubmission V1 (currently deployed) uses
+> announce-to-all with deduplication, which propagates each tx body along a
+> spanning tree (~1 effective send per node). CIP-164 simulations use V1-style
+> gossip and give ~2× lower tx egress; V2 is the target protocol for both
+> Praos and Leios.
```
ch1bo: @nfrisby @coot How wrong are these assumptions?

docs/cost-estimate/04-egress.md
```diff
@@ -21,17 +21,22 @@ The Cardano relay network is asymmetric. A typical relay node has:
...
 > [!Note]
 >
-> The block/EB body peer count (4) derives from: each of 100 inbound peers has
-> ~25 upstream sources and requests from ~1 of them on average → 100 × 1/25 = 4.
+> The block/EB body peer count derives from: each of 100 inbound peers has ~25
+> upstream sources and requests from M of them simultaneously (fetch multiplicity
+> M). Our expected egress peers = 100 × M/25. For a lean relay M=1 → **4 peers**;
+> for a beefy relay fetching from M=4–8 peers to reduce tail latency → **16–32
+> peers**. High-stake pool relays commonly use M=4–8 to minimise block arrival
+> latency at the cost of higher egress.
```
ch1bo, line 38: @nfrisby This is what we talked about today and what you recalled @karknu say in the past. Please correct me if this is a wrong summary here

ch1bo:
Merging this already so I can point people to these new numbers - these are definitely more representative of what we are currently implementing (CIP-164). Anyone please feel free to double check what I pointed at still.

docs/cost-estimate/04-egress.md
nfrisby: I think that text matches my new understanding

docs/cost-estimate/04-egress.md
nfrisby: I defer

docs/cost-estimate/02-compute-ram.md
nfrisby: What's the "1,000  B/s" term in "TxkB/s × 1,000  B/s × 3,600  s"?

My argument about the necessary TxCache size in the [February public monthly review presentation](https://docs.google.com/presentation/d/1oLUDt3rEDLwkwQyvRED13nEwQ6fcogg5X4SWlsCcj7Q/edit?slide=id.g3cae8abedd7_0_147#slide=id.g3cae8abedd7_0_147) was based on a couple choices: worst-case density of elections X and also how little stake Y some fragmented mempool needed to have in order to benefit from the TxCache.

As Y gets small, data needs to stay in the TxCache longer, so that it'll still be there the next time that Mempool issues an EB.

As X gets big (ie you lower your tolerance for cache misses), there's potentially more fresh data flowing into the TxCache, so stuff will be evicted more quickly---higher flux.

So decreasing Y (ie helping not just the big mempools overcome fragmentation) and rising X (ie higher success chance) are in tension; ie both require making the TxCache even bigger.

For my presentation, I chose something like 5-10% stake for the Mempools we expect the TxCache to help with; for ~99% or ~99.9% success rate, that meant the TxCache needed to retain things for ~15 minutes. And then 99.99% of the time, we'd at most ~65 elections in a 15 minute window. So the TxCache would need to accommodate ~65 EBs (all potentiall full and no overlap) to "almost always" still remember a 5-10% Mempool's previous EB when it's next EB arrived.

I think we can round that up to ~100. So that's 1.25 GB of EbBody+EbClosure (pessimistically assuming 0% overlap). The role of the TxCache also requires indexing ~140 million txs themselves (not just the EbHash), so that's a nontrivial eg hash table (multiple 100s of MB) too, if that index needs to live in memory for quick access during the LeiosFetch decision loop, eg.

PR #880, opened 2026-04-29
Add Mininet LeiosFetch logic test bed to targeted-investigations (Frisby April 2026)

### LINKED ISSUE #853: Design fetch logic with bounded resource usage and bounded EB diffusion tail latency (April)
### Why

This Issue is a rotation of https://github.com/input-output-hk/ouroboros-leios/issues/811 from March into April. These Issues track a design/exploration effort on a monthly basis; we can only broadly anticipate their content at the beginning of the month.

- The Leios Fetch Logic in the existing prototype has poor resource usage bounds. Especially in the worst-case, it risks allocating hundreds of millions of heap objects, with gigabytes worth of size. It's fine for the demos so far, but must be refined for a proper prototype node.
- Moreover, the Rust simulations used to determine the feasible message delay upper bounds recommended in the first version of the CIP-0164 were derived using a very abstracted network stack. That same logic is not realistic/mainnet ready.

### What

This task involves determining how to implement a concrete decision logic that can provide similar behaviors as the simulations underlying the CIP-0164 "feasible" parameter values while also avoiding excessive resource usage and excessive tail latency, even in the worst-case. If that perfect robustness is not possible, we at least need a good argument for why it will fail sufficiently _in_frequently.

### How

Co-locate the known challenges, compare and contrast with prior art (e.g. other high through-put blockchains), develop a rough draft proposal, discuss with other experts and architects (eg Networking Team). Iterate. Likely to ultimately involve addition simulations, since the emprics of the topology itself are so impactful and because "worst-cases" are difficult to exercise outside of simulations (e.g. without arbitrary EB injection/at smaller scales/etc/etc).

nfrisby:
Resolves https://github.com/input-output-hk/ouroboros-leios/issues/853

This PR checkpoints my working material developed during April 2026. The ultimate target of this work is a set of rules that controls which requests the LeiosFetch decision logic issues, to which nodes, when. This PR doesn't answer that question, but it lays the groundwork for assessing the different possibilities and their feasibility.

### About LLM

This PR introduces a lot of text and code. It's my first project that relied heavily on Claude, so that's not particularly surprising. "Greenfield" exploration features explosive growth, and Claude only accelerates that.

For what it's worth: I scrutinized all of the diffs Claude generated, except for plotting---if the rendered plot looked right I didn't personally scrutinize it (though I sometimes asked Claude to double-check). That said, plenty of this stuff is in domains I'm not particularly expert in to begin with, so eventually deserves another set of eyes (eg low-level socket syscalls, the Linux tc netns API, etc). But at least every diff looked plausible and relevant to the prompts I has issued.


nfrisby:
Rendered entrypoint: https://github.com/input-output-hk/ouroboros-leios/blob/nfrisby/april2026-checkpoint/docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/README.md

This targeted investigation avoids the use of Leios jargon for a few reasons. Principally: it discusses  "payloads" and "payload components" instead of eg "EbClosures" and "transactions".

1) This makes it more approachable for non-Leios experts (e.g. TCP experts, networking experts, etc).
2) The relevant prior art is a wide net, and it uses more generic terminology.
3) I wanted Claude/Gemini Pro/etc to engage with the more general terms rather than Leios jargon; it seemed wise to not rely on them to generalize my wording.

nfrisby:
Based on our team's discussion last week, I studied TCP CUBIC's rules this week and steered Claude into authoring this short memo/worked example.

https://github.com/input-output-hk/ouroboros-leios/blob/nfrisby/april2026-checkpoint/docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/loss-model-memo.md

@coot @crocodile-dentist @karknu Does it look coherent do you? Is it overlooking other factors / options? Thanks.

nfrisby:
Based on my 1:1 with Sebastian on Friday May 8th, I was thinking "outside the box". This new sketch is the result. I don't plan to pursue this much more, unless the team changes our mind about adaptive attack vectors.

https://github.com/input-output-hk/ouroboros-leios/blob/nfrisby/april2026-checkpoint/docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-EbClosureMinimumAge.md

(Edit: I think it's defensible on this PR because "The Risk" is the exact kind of thing this testbed is supposed to help us understand. But we'd probably relocate it if the team got more interested in it.)

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+    - Never have more than X outstanding bytes of requests sent to a single low-latency peer (TODO tune X).
+    - Only start sending overlapped requests among low-latency peers once there are ≤X bytes of the payload missing---i.e. once the non-overlapped requests could no longer saturate X.
+- For high-latency peers:
+    - Immediately request all of the still-missing portion of the payload from the first high-latency peer to offer it.
+    - Don't request any part of that payload from another high-latency peer until the first request times out.
+      (TODO is that timeout necessarily fatal?
+      If it's not fatal, then the network resources (including the kernel buffers) might not actually be free yet… but at least the application-level buffers would already be free?)
+    - Note that we need to send the request even if *some* of the data has already arrived from low-latency peers---the high-latency peer might be the only one that will actually send _every_ part of the payload (i.e. might be the only honest peer).
+
+## Conclusion
```
nfrisby, line 85: @ch1bo @pagio @coot @crocodile-dentist @karknu

This idea occurred to me on my plane ride this morning and seems very promising! I wrote up this document to introduce it to the Leios group. What do you think?

As per usual, this idea seems somewhat obvious in hindsight, but I hadn't before appreciated this particular implication of the "sweetspot" in the Venn diagram of these two observations (1. can use low-latency peers efficiently and 2. using every low-latency peer promptly justifies not _necessarily_ using more than one high-latency peer promptly, since it effectively shares high-latency peers within the neighborhood).

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
+## Introduction
+
+This memo proposes some requirements for the LeiosFetch decision logic, derived from the design of Cardano's P2P mechanisms.
+
+The new insight that motivates this document is that the intended network dynamics across the topology inform not only the algorithm for selecting peers but also the algorithm for allocating resources among the selected peers.
+
+## Observation 1
+
+Cardano architects leveraged technical reports from [Spyros Voulgaris](https://acropolis.aueb.gr/~spyros/) et al of Athens University of Economics and Business when designing the P2P peer sharing and selection algorithm.
+A key recommendation from these reports was that a significant fraction of a node's peers should be low latency and another significant fraction should be high latency (or random instead seems to suffice).
```
crocodile-dentist: My interpretation of the report is that explicitly picking high latency peers is a hazard, so to avoid misunderstanding I would point out that the latter is the intended outcome but not a premeditated  general strategy.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+The intended dynamics are for intercontinental peers to first deliver a message between distant neighborhoods, and then the low latency connections would ensure the message rapidly spreads within each neighborhood.
```
crocodile-dentist, line 12: I guess it is not incorrect, but I see the dynamics or the practical aspect differently, esp. since we do not explicitly filter for high latency/intercontinental peers so there isn't a direct mechanism of fast long distance transfer - at least as far my understanding of the report goes. My interpretation is that each node has a relatively sparse/small valency (eg. 10C10R) so to 'ensure' prompt long distance transfer we want to grow our local cluster with some fraction of low rtt peers, such that this cluster as a whole has a high probability of direct connection to far away peer(s) by randomness alone / the random fraction of peers at each relay. The latter achieves the intended outcome for the cluster.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Do not send redundant requests (a.k.a. _hedge requests_) to high-latency peers.
+- A request to a high-latency peer should request as much (useful) data as possible (e.g. all missing transactions from the EB closure).
+  This ensures the sender will utilize as much of the link's bandwidth capacity as possible (i.e. up to the large Bandwith-Delay Product/Mathis ceiling/etc) instead of being unnecessarily application-limited.
+- A request to a high-latency peer must have a relatively generous timeout, something like 1 second.
+  Once that timeout fires, the node should switch its requests to a different high-latency peer.
+- In contrast, requests to a low-latency peer should request a limited cumulative amount of data (and hence resource utilization) and a tight timeout.
+  This allows a decomposable payload (like a EB closure) to be split among requests to multiple low-latency peers simultaneously.
+  Only once the set of still-missing components is small enough for a single request would it be useful to send redundant requests to multiple peers, in order to prevent an adversarial peer from slow lorising the final missing piece.
```
crocodile-dentist, line 41: @karknu 's work on tx submission shows that hedge requests with small timeouts have promising performance. Maybe we shouldn't wait too long before issuing redundant requests even for our 'nearby' peers, ie. waiting until still-missing components is small enough for a single request.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+    - If one node in the neighborhood _doesn't_ have that attacker as peer, then the risk is already mitigated.
+      How to determine if that's already sufficient likely?
+      If it is, then every node could simply request the payload from the first high-latency peer to offer it.
+- TODO We're tacitly assuming that it's not worth the complexity/inefficiency for LeiosFetch to ensure that even a node that doesn't have any honest low-latency peers will receive EB bodies and closures promptly.
+
+TODO so the optimistic logic is:
+
+- Consider low-latency and high-latency peers independently.
+    - (The only interaction among low- and high-latency peers is that data that has already _arrived_ from one doesn't need to be requested from the other.)
+    - TODO how to assess whether a peer is low- or high-latency?
```
coot, line 72: We should design a trace and get some real data to see what are the latency distributions, and how stable they are etc.  Likely peers are grouped based on the continents then we can choose some sensible quantile or something more sophisticated.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+  Only once the set of still-missing components is small enough for a single request would it be useful to send redundant requests to multiple peers, in order to prevent an adversarial peer from slow lorising the final missing piece.
+  As a result, only a small amount of data would be requested more than once among low-latency peers.
+
+**Risk**.
+If the adversary can increase the probability of it being the first high-latency peer to offer some datum to all nodes in a neighborhood, then they could be a slow loris such that _none_ of the nodes in that neighborhood receive the datum from a high-latency peer.
+The whole neighborhood would be delayed by at least one timeout and potentially more.
+
+If that risk is unarmed, then some node in the neighborhood will receive the payload from a high-latency peer, and then it will spread quickly among the neighborhood.
+It's as if all of the nodes in the neighborhood shared their collective high-latency peers.
+So, unless the risk above is armed, each neighborhood can only be deprived of messages originating outside the neighborhood if _all_ of the nodes in that neighborhood happen to choose a slow loris peer.
```
crocodile-dentist: Maybe we can make an even stronger claim, that the attacked would have to be a slow loris to every far away cluster. If he is just a slow loris to cluster A, but not B, then A can get the attacker's messages thru B, which may be faster than getting it from him directly. If his goal is to somehow split the network or cause disruptions, this requirement would make this tactic less effective?

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+
+The key ideas are as follows.
+
+- Utilize multiple low-latency peers simultaneously, with a relatively small amount of outstanding data for each peer at any given time.
+- Send overlapping requests to multiple low-latency peers only when constructing enough requests to saturate a low-latency peer would otherwise be impossible.
```
crocodile-dentist, line 56: This sounds appealing, but it would require us (network) knowing how much performance/bandwidth we can squeeze out of a peer apriori. Maybe add this as a note?

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+- Do not utilize multiple high-latency peers simultaneously---their dynamics require large requests and so utilizing multiple would require requests to overlap too much.
+- TODO Somehow avoid making it too easy for an attacker to arrange for every node in a neighborhood to choose a slow loris attacker as the one high-latency peer they utilize.
```
crocodile-dentist, line 58: The larger the local cluster, the less likely every peer would choose the same slow loris so maybe we can sweep this under our P2P valency argument that each peer utilizes a few random peers?

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
crocodile-dentist, line 72: To make this portable, it would require some cooperation between peers.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+      If it is, then every node could simply request the payload from the first high-latency peer to offer it.
+- TODO We're tacitly assuming that it's not worth the complexity/inefficiency for LeiosFetch to ensure that even a node that doesn't have any honest low-latency peers will receive EB bodies and closures promptly.
+
+TODO so the optimistic logic is:
+
+- Consider low-latency and high-latency peers independently.
+    - (The only interaction among low- and high-latency peers is that data that has already _arrived_ from one doesn't need to be requested from the other.)
+    - TODO how to assess whether a peer is low- or high-latency?
+      Just some (bandwidth-independent?) (load independent?) RTT cutoff Y (TODO tune Y)?
+    - TODO does that assessment need to change over time?
```
crocodile-dentist, line 74: I would say so, esp. for long distance peers where congestion and routing changes and other anomalies may be more probable than for short jumps.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+- For low-latency peers:
+    - Never have more than X outstanding bytes of requests sent to a single low-latency peer (TODO tune X).
+    - Only start sending overlapped requests among low-latency peers once there are ≤X bytes of the payload missing---i.e. once the non-overlapped requests could no longer saturate X.
+- For high-latency peers:
+    - Immediately request all of the still-missing portion of the payload from the first high-latency peer to offer it.
+    - Don't request any part of that payload from another high-latency peer until the first request times out.
+      (TODO is that timeout necessarily fatal?
```
crocodile-dentist, line 81: As in drops the connection? Unless it is large enough to trip our protocol limits, my hunch is to say no. Maybe we can introduce some scoring metric?

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
...
+- A request to a high-latency peer must have a relatively generous timeout, something like 1 second.
+  Once that timeout fires, the node should switch its requests to a different high-latency peer.
+- In contrast, requests to a low-latency peer should request a limited cumulative amount of data (and hence resource utilization) and a tight timeout.
+  This allows a decomposable payload (like a EB closure) to be split among requests to multiple low-latency peers simultaneously.
+  Only once the set of still-missing components is small enough for a single request would it be useful to send redundant requests to multiple peers, in order to prevent an adversarial peer from slow lorising the final missing piece.
+  As a result, only a small amount of data would be requested more than once among low-latency peers.
+
+**Risk**.
+If the adversary can increase the probability of it being the first high-latency peer to offer some datum to all nodes in a neighborhood, then they could be a slow loris such that _none_ of the nodes in that neighborhood receive the datum from a high-latency peer.
```
pagio, line 45: The adversary can claim to have the block without really having it, and thus delay everyone by one timeout

pagio: Maybe that is the worse it can do.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,83 @@
+## Introduction
+
+This memo proposes some requirements for the LeiosFetch decision logic, derived from the design of Cardano's P2P mechanisms.
+
+The new insight that motivates this document is that the intended network dynamics across the topology inform not only the algorithm for selecting peers but also the algorithm for allocating resources among the selected peers.
+
+## Observation 1
+
+Cardano architects leveraged technical reports from [Spyros Voulgaris](https://acropolis.aueb.gr/~spyros/) et al of Athens University of Economics and Business when designing the P2P peer sharing and selection algorithm.
```
nfrisby, line 9: Karl has called this sentence out as an overstatement. TODO back it off

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: I'm working on a write-up that dives further into the details along these lines. Sketch so far:

1) bound the inflight bytes to X for each low-latency peer, in order to bound the amount of resources (eg socket buffers, queues in paths' bottlenecks, etc) they could collectively use
2) keep X inflight with all of your low-latency peer
3) eventually you can't keep X inflight with a low-latency peer unless you start sending overlapping requests; but don't introduce any overlap before you have to (ie once there's <X bytes left among the EB's transactions that you haven't already requested)

So it would involve "hedge requests", but no sooner than they'd be useful: you should prefer a non-hedge request when such a thing would saturate your resource budget and only fall back to hedge requests if not.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: I'm working through the arithmetic on this now. I'm hoping something "dumb" will work. Something like "if a peer has latency below 75 ms, then limit their inflight requests to 120 kB." The key point: this is regardless of their observed goodput.

I'm working up a calculator to explore that space.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: I added the second half of this as a direct quote. Thanks!

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: I agree with what you wrote.

In my text, I was thinking in terms of whether the neighborhood receives the message "promptly", but I didn't make that explicit. I probably just tired of writing "promptly" :/

Receiving the A message via B would be slower than receiving it from an honest member of A. But definitely still better than _never_ receiving it, yeah.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: I'm going to reword this. By "saturate" I meant the "relatively small amount of outstanding data" from the previous bullet point, which I'm hoping can be "good enough" even if it underutilizes the specific link's goodput capacity.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: I agree; that's what I elaborated on somewhat below with.

> If one node in the neighborhood _doesn't_ have that attacker as peer, then the risk is already mitigated.

But if all the nodes in the neighborhood do have that attacker as a high-latency peer, then nothing prevents the attacker from "offering" the block even before they have it, which might be enough to cause the whole neighborhood to allocate their high-latency request budget to that peer and so ignore offers from honest peers for some duration (ie until the initial request timed out).

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: yeah, fatal = disconnect. Made that explicit.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: I made the text more explicit.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
nfrisby: Rendered: https://github.com/input-output-hk/ouroboros-leios/blob/nfrisby/april2026-checkpoint/docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
ch1bo: The adversary would risk a disconnect though if they can't serve the request following an offer.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
ch1bo: > from "offering" the block even before they have it,

Wouldn't the attacker risk their network position if they do this? i.e. nodes would disconnect if the adversarial node can't serve what they offer.

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
```diff
@@ -0,0 +1,152 @@
...
+
+- Utilize multiple low-latency peers simultaneously, with a relatively small amount of outstanding data for each peer at any given time.
+- Send overlapping requests to multiple low-latency peers only when constructing enough requests to saturate a low-latency peer would otherwise be impossible.
+- Do not utilize multiple high-latency peers simultaneously---their dynamics require large requests and so utilizing multiple would require requests to overlap too much.
+- TODO Somehow avoid making it too easy for an attacker to arrange for every node in a neighborhood to choose a slow loris attacker as the one high-latency peer they utilize.
+  For example, the slow loris attacker could arrange to be the first high-latency peer to offer the block (e.g. even before they have it).
+    - On the one hand: you want to request the block from the first peer that offers it to you; if they're honest, then it's just extra latency to not immediately request it.
+    - But the adversary can arrange to be that peer.
+    - So some nodes would have to (randomly) wait for additional high-latency peers to offer them the same payload, and only react to the second (or even third) offer?
```
ch1bo, line 62: I think we should hedge for this scenario by always requesting
- Everything from first-to-offer high-latency peer
- Everything from currently-best of the high-latency peers, once they offer
- ... partial requests according to your low-latency peer semantics

docs/targeted-design-investigations/mininet-LeiosFetch-test-bed/IDEA-NeighborhoodFarPeers.md
ch1bo: I like the idea. Especially as it leverages the apparent topology of the Cardano network. (Can we assume the topology does not change? Or how would we monitor? Or how to make future network designers aware on our assumption?)

Is it possible though? It depends on having a continued understanding of at least low vs. high latency peers, and maybe even a more granular ordering by latency - how can this be attained?

crocodile-dentist: > Is it possible though? It depends on having a continued understanding of at least low vs. high latency peers, and maybe even a more granular ordering by latency - how can this be attained?

I'm working on expanding the muxer to expose an estimate of rtt for client application use so this line of work here can be supported.

PR #885, opened 2026-05-06
Leios design: Write down some guidelines on the cardano-node architecture

ch1bo:
These are deliberately vague and not fully fleshed out: I'm still building my opinion about the whole consensus codebase and how abstract/concrete code works out there. But I wanted to capture already some high-level ideas of being as concrete as possible and have little conditional wiring... and get feedback on these!

This was done in the context of exploring various designs for the voting thread only enabling in a leios-compatible era: https://github.com/IntersectMBO/ouroboros-consensus/blob/abd1c00ccec736c9155db8d6f9d83b2e2106ece6/leios-voting-design.md


docs/leios-design/README.md
```diff
@@ -441,6 +441,26 @@ While being a significant change to the consensus protocol, Leios does not requi
 >
 > TODO: Explain why focus on relay node (upstream/downstream relationship); briefly mention block producer node differences; Add similar diagram for block producer? block and vote production not shown in relay diagram
 
+### Design guidelines
+
+The guidelines below inform how the new components in the diagram are structured internally and how they integrate with the existing consensus codebase. They are not Leios-specific, but articulate a discipline that consensus extensions in general could follow.
+
+#### Avoid conditional wiring
```
nfrisby, line 448: 1000% happy with this guideline 👏

docs/leios-design/README.md
```diff
@@ -441,6 +441,26 @@ While being a significant change to the consensus protocol, Leios does not requi
...
+
+#### Be as concrete as possible
+
+Use concrete types mono-morphic types where possible; only add a parameter when today's code already needs the alternative. The clearest established example is `m`: consensus code is parametric over `m` because `IO` runs the production node and `IOSim` runs deterministic tests — both committed inhabitants, exercised on every CI run. Everything else stays concrete by default: `LeiosVote` is a BLS-signed record, `Committee`, `Round`, and `EB` are concrete data types, and component handles like `LeiosVoteState m` and `LeiosDbHandle m` are parametric only over `m`. Where consensus already imposes polymorphism — the `blk` block format or `LedgerState blk` at the ledger boundary — a single typeclass or record of accessors consumes `blk` and projects to those concrete types, so the polymorphism does not leak further. When a future consumer is committed (for example, tests that need a fast deterministic signature scheme rather than running real BLS under IOSim), the corresponding parameter is added at the point the second inhabitant exists, not in anticipation.
```
nfrisby, line 458: "concrete types mono-morphic types" seems like a typo

docs/leios-design/README.md
```diff
@@ -441,6 +441,26 @@ While being a significant change to the consensus protocol, Leios does not requi
...
+
+The trade-off is a steady-state cost: idle threads parked on STM, empty caches, and registered-but-quiet channels on every long-lived peer connection — paid universally, including by nodes that sync only historical chains and will never see Leios active. We accept it in exchange for a single source of truth and a hard fork that is purely a ledger event, not a code path. The Hard Fork Combinator already requires us to keep present-but-inactive behavior for past and future eras, and the guideline leans into that: Leios is enabled by entering the Leios ledger era and by the corresponding protocol parameters — set through genesis files or on-chain governance like any other parameter — not by a separate runtime feature flag. The one exception is NodeToNode protocol versioning, which is genuinely a code-side decision: shipping a new mini-protocol requires a code release, so the version that introduces the Leios mini-protocols is negotiated at handshake rather than read from the ledger. The guideline still rejects per-component, per-thread, per-protocol branching that would recreate feature awareness throughout the codebase.
+
+#### Be as concrete as possible
```
nfrisby, line 456: I also like this guideline, but I'm suspicious of the ...

> no [...] `era` parameters propagated through Leios types without a [...] boundary that actually uses them

... part. So I suspect there's some murkiness being papered-over by the current text of this paragraph in the PR.

For example, I'm extremely confident that the header codec and the tx codec will change during future hard forks. Since some of our Leios mini protocol messages contain headers and txs, I suggest pre-emptively including the HFC-envelope in those payloads' codec---even if we don't use the HFC itself in the implementation yet. Just include an "HFC era tag" somewhere in the message's CBOR codec. Its value should be `7` for Dijkstra (if I recall correctly).

What's the cost of not doing so in the first implementation? It's the obligation of then managing an extra backwards-incompatible version bump of the mini protocol itself (negotiated at the peer-to-peer handshake). If the codec needs to change because Leios itself changed, then we might have to introduce a new mini protocol version anyway. But if the Leios mini protocols would need to add a new major version merely because an "unrelated" hard fork changed some codec that Leios messages happen to carry (eg an RB header), then that seems undesirable (especially from the release engineer's perspective!).

Is avoiding that laborious task worth the overhead of including an HFC-envelope now? I think so, at least for things like RB headers. It's a very minor overhead. Especially if you consider the fact that multiple node implementations' devs might need to jump through that hoop the first time the need arises.

The simple cautious option would be: just add an era tag to every message that has a non-trivial payload. Maybe you're never going to need it. Sometimes you will, and adding this one simple field avoided some obnoxious error-prone work that multiple teams might have had to separately replicate. (None of this is _on chain_; it just the wire protocol I'm talking about.) The deepest overhead of this option is that now every component that creates a message needs to know which era tag to put in it. That information should always be available (eg it's true today for ChainSync and BlockFetch), but maybe it's possible some component could have gotten by in the long-term without managing it 🤷 

-----

However, in the case of the _specific Leios messages_ that contain transactions, I could also imagine that we'll never need that HFC era tag for those messages, even if hard forks do change the tx's codec: at least _the diffusion part_ of EBs should just be slinging around bytestrings, only checking hashes instead of doing full decoding. There are only a few components of the Consensus Layer that would actually need to decode/encode these bytestrings _as transactions_, and those components will have enough context to know which era's codec to use (since they know what era the EB was announced in). So even though "transaction codecs can change at hard fork era transitions" and "`MsgLeiosBlockTxs` carries transactions", maybe we wouldn't benefit from `MsgLeiosBlockTxs` including the HFC era tag?

... On the other hand, it wouldn't hurt to include it anyway. Also, maybe a developer doing debugger work might appreciate having the era tag at hand, even though it's technically redundant?

For the vote-carrying messages, if those need to change (eg new committee selection/aggregation scheme/etc), then it's probably Leios itself motivating that change, and so a N2N major version bump might already be necessary. Something as sophisticated as adding Forward Erasure Codes would definitely need both a N2N major version bump and a hard fork.

In these cases, I'm more sympathetic to the YAGNI idea. Don't try to predict what you need, and just deal with the labor of bumping the mini protocol major version once you have enough information to know what you actually need. But for some messages, I'm pretty sure You *ARE* Gonna Need It (eg the RB header in the EbAnnouncement), so we might as well add a tiny datum that could avoid the obnoxious multi-team backwards-compatibility dance later.

The mixed feelings I have that are apparent above are already captured by these sentences...

> The line between "present need" and "anticipated need" is judgment, not a rule; a present need six weeks away is still present. When introducing polymorphism, the question to answer in the PR description is _which current consumer needs the alternative inhabitant?_ If the answer is "none yet," the answer is "delete the parameter."

... but something about them reads to me in a way that seems over-simplified---I can't precisely say, though... maybe the emphasis on "current"?

ch1bo: Thanks for adding this perspective! I have not considered wire formats so far and this sounds more like a "version your encoded data" kind of guideline? Happy to have that though!

In terms of staying concrete where possible this would be compatible too: Either 1. the thing we encode is already abstract, e.g. `Header blk` would contain the HFC tag (somehow) and the concrete encoder can just include it; or
2. the thing to encode has only one inhabitant and it should not hurt to just put a "V1" equivalent into any encoding of that value

I think versioning encoded (for the wire, persisted, etc.) data is a good idea, although I have been also in the camp of "no version is version 0" in the past. But that is more like how the decoders need to handle this (i.e. we would be predicting the needs of versioning)

I certainly should update the text to be more clear about this. Should I capture the versioning encoded data as a separate guideline?

nfrisby: The nub of my top comment is "we have a precedent that some message's payloads are versioned via the HFC's era tag, and so we could avoid some onerous work/delays later by already having the first implementation already include the HFC era tag in its on-the-wire codec, even if the first implementation doesn't actually otherwise leverage the HFC itself".

I agree with your either 1 or 2 characterization. Option 1 comes for free if the impl actually does use the HFC under the hood. Option 2 is the alternative---and my original comment went further on to suggest that Option 2 could be implemented in such way that it's actually indistinguishable (from the peers' perspective) from Option 1. And the HFC has already been established as a sufficient mechanism for managing on-the-wire codecs changing due to hard forks across era boundaries.

> although I have been also in the camp of "no version is version 0" in the past

The Byron-to-Shelley work had to cope with that approach, and it turned out to be a pain in the ass---today's HFC has some cruft that would have be unnecessary if its responsibilities had been anticipated back when the Byron codec was chosen.

> Should I capture the versioning encoded data as a separate guideline?

That seems like a promising way to organize it yeah.

However, I also think the "Be as concrete as possible" section should be less vague about the trade-offs. It said "is judgment, not a rule", but as-written the whole thing just smells like a rule. Demonstrating some of the mentioned judgment would appeal to me. In particular, discussing what the actual trade-offs might look like. Sometimes you'll see that a little extra work creates the possibility of avoiding more extra work later. Judging the Expected Value there is tough without considering concrete details.

In my comment, I spelled out the risk/reward on either side. And it's a healthy chunk of text, so it makes sense to isolate it. But I do think this "Be as concrete as possible" section would then need to explicitly reference the "on-the-wire versioning" discussion as an example how how to considered the trade-offs when exercising the "judgement" it mentions.

PR #886, opened 2026-05-08
filip(fix): hide 2 old homepage sections

fstoqnov-iohk:


olgahryniuk:
@ch1bo, as agreed earlier today, we don't need the how it works and research explainers on the landing page. Removing these avoids repetition and ensures seamless flow through the tracker.
This copy nicely lives in our introduction here: https://leios.cardano-scaling.org/docs/what-is-leios#from-research-to-reality  
Thank you!

PR #887, opened 2026-05-11
Split the monolithic cabal.project into per-cluster Haskell projects

ch1bo:
Each Haskell cluster (simulation, leios-trace-hs, leios-trace-verifier, trace-processor, leios-deltaq, betti0) now owns its own cabal.project and its own haskell.nix project scoped to just its subtree, exposed lazily under legacyPackages.<cluster>. The previous setup snapshot the whole repo as haskell.nix source, so any edit invalidated every cluster's plan and IFD ran on every nix develop; now eval inputs are bounded per cluster, dev shells are plain mkShell decoupled from haskell.nix (~10s entry), and clusters can drift their pins independently.

Cross-cluster source deps (simulation, leios-trace-verifier both consume leios-trace-hs) use lib.fileset.toSource + cabalProjectFileName so the existing ../leios-trace-hs paths resolve naturally. iogx is dropped; flake-parts drives the flake directly with haskell-nix and CHaP as top-level inputs. CI now scopes to the simulation cluster.


ch1bo:
@copilot resolve the merge conflicts in this pull request

PR #889, opened 2026-05-12
Add improved DeltaQ analysis for Linear Leios EB diffusion

crocodile-dentist:
Fixes three bugs in the prior Haskell analysis and adds a parametric sweep over EB closure size (S_EB_tx) to determine practical limits.

Adds a sensitivity study to analyze effects of mempool fragmentation: 
1) 1-hop vs. multi hop closure fetch:
   a. Full-blended (12MB) worst case
   b. Maximum feasible closure size when 1-hop fails.
2) TxCache miss rate sensitivity study for 1-hop missing tx's closure fetch

Key corrections:
- RB structure: certRB and txRB paths are mutually exclusive (probabilistic choice weighted by p_cert), not forced in parallel as in the prior model
- Scale mixture → fixed-N CLT: EB closure validator processes all N txs, not a random subset; prior model underestimated reapply time by 2×
- apply vs. reapply: cache-miss txs (π₁=1/6) require full applyTx (0.507 ms/tx), not just reapplyTx (0.070 ms/tx); effective cost μ_eff = 0.143 ms/tx
- Adds two TCP congestion control algorithms to simulate transfer times from cold start and with steady state max thruput given a set loss parameter.

Lists assumptions, caveats, and items which deserve further study.

[report.md](https://github.com/input-output-hk/ouroboros-leios/blob/88e01bd0301ab6d9846cd17b07916745b2c017ed/analysis/deltaq/improved-leios/report.md)


analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+
+---
+
+## 2. Protocol Recap
+
+Linear Leios extends Ouroboros Praos by adding Endorser Blocks (EBs). The
+relevant timing pipeline (one round) is:
+
+```
+  slot:  |---L_hdr---|---L_hdr---|---L_vote---|---L_diff---|---L_hdr---|
```
yveshauser: Why has the first L_hdr been moved to end of the sequence? 
```suggestion
  slot:  |---L_hdr---|---L_hdr---|---L_hdr---|---L_vote---|---L_diff---|
```

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+
+---
+
+## 3. Bugs Fixed in the Prior Analysis
+
+### 3.1 RB path structure
+
+Original code (`DeltaQ/Leios.hs`, line 128):
+```haskell
+processRBandEB = processRB ./\. processEB   -- WRONG: forces both paths
```
yveshauser, line 187: Yes, as EB is optional, `processEB` is optional. But when there is an EB, it needs to be processed, certification may fail. Therefore we should use the probability of having an EB rather than having a certificate when using the `choice` operator

analysis/deltaq/improved-leios/analysis.py
yveshauser, line 1: Running the analysis completes almost instantly, nice! We could put it all into a Jupyter notebook for interactive exploration

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+$$N_{txs} = \min\left(\lfloor S_{EB-tx} / \overline{t_x} \rfloor,\; 16000\right)$$
+
+where $\overline{t_x}$ = 1 kB (average transaction size) and 16 000 is the
+maximum number of transaction references a 512 kB EB body can hold.
+
+### 3.3 Scale mixture vs. fixed-N distribution for reapplication {#scale-mixture}
+
+The original model used a *scale mixture* for the reapply CDF:
+
+$$F_\text{mix}(t) = \frac{1}{N}\sum_{k=1}^{N} \Phi\!\left(\frac{t - k\mu}{\sqrt{k}\,\sigma}\right)$$
```
yveshauser: `!` does not belong here, does it?   
```suggestion
$$F_\text{mix}(t) = \frac{1}{N}\sum_{k=1}^{N} \Phi\left(\frac{t - k\mu}{\sqrt{k}\,\sigma}\right)$$
```

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+
+This models a block of *random* size $k \sim \mathcal{U}(1, N)$ – appropriate
+for the Praos `applyTxs` path where the block size is unknown, but **wrong** for
+EB closure validation where the validator must process *all* $N$ transactions.
+
+Fixed: use the CLT approximation for the *fixed-N* case:
+
+$$F_\text{fixed}(t) = \Phi\!\left(\frac{t - N\mu}{\sqrt{N}\,\sigma}\right)$$
```
yveshauser: Same as above
```suggestion
$$F_\text{fixed}(t) = \Phi\left(\frac{t - N\mu}{\sqrt{N}\,\sigma}\right)$$
```

yveshauser (approved):
Thanks for producing this report. I reviewed it with a focus on the proposed model changes compared to the previous report. I agree to the corrections
 
- deriving apply / reapply as weighted arithmetic mean from the files (4.3)
- replacing the scaled mixture distribution with a fixed size distribution (3.3)

After merging the PR, my plan is to produce a Jupyter notebook out of it, in order to facilitate exploring the numbers interactively

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+network-induced sensitivity.
+
+*Why TCP congestion control matters here.*  Cardano peer-to-peer diffusion of
+blocks, EBs, and EB-closure transactions runs over long-lived TCP connections
+between SPO nodes.  Effective throughput on each connection is determined not
+by raw link capacity but by the **TCP congestion-control algorithm** running
+in the sender's kernel.  Reno (RFC 5681, the legacy AIMD reference) and CUBIC
+(RFC 8312, the modern Linux default since 2.6.19) respond differently to
+packet loss: their steady-state throughput scales as $p^{-1/2}$ (Reno/Mathis)
+vs $p^{-3/4}$ (CUBIC), giving CUBIC ~11× more throughput at $p = 10^{-4}$.
```
RandallYe, line 61: It might be good to briefly mention why $p = 10^{-4}$ is chosen.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+| 8 MB        | **0.497**                | **0.497**               | —      |
+| **12 MB**   | **0.497**                | **0.497**               | —      |
+
+Under **Mathis** ($T \propto p^{-1/2}$, the conservative Reno/AIMD model), the
+2 MB missing-closure fetch at 12 MB takes ≈ 4.4 s over long-haul (134 ms OWD →
+268 ms RTT) links — comfortably within the 7 s voter deadline.
+$P_\text{validating} = 0.948$, $P_\text{quorum} \approx 1.000$, and
+$P_\text{cert}$ stays at the Praos cap of ≈ 0.497.
+
+Under **CUBIC** ($T \propto p^{-3/4}$, RFC 8312), the same fetch completes in
```
RandallYe: RFC 8312 is already mentioned before and it could be ignored here.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+prior-model default $\pi_1 = 1/6$, RTTs taken from the original Praos paper's
+Table 1; see §5.5 for the empirical $\pi_1 \approx 0.06$ update):**
+
+CPU is not the binding constraint at any closure size ≤ 12 MB.  Single-core
+voter CPU completes in $< 2$ s for the 12 MB case ($\mu_\text{eff} = 0.143$
+ms/tx).  The **1-hop missing-closure fetch** is the dominant network step,
+and it completes well within the 7 s voter deadline at every CIP-target size
+under either throughput model:
+
+| $S_{EB-tx}$ | $P_\text{cert}$ (Mathis) | $P_\text{cert}$ (CUBIC) | Δ      |
```
RandallYe: This table has all the $P_{cert}$ at 0.497. It could be just one sentence to summarise it if you would like to make it compact.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+
+Under **CUBIC** ($T \propto p^{-3/4}$, RFC 8312), the same fetch completes in
+≈ 2.0 s.  $P_\text{validating} = 0.992$, $P_\text{quorum} \approx 1.000$, and
+$P_\text{cert}$ is also at ≈ 0.497.
+
+**Bottom line.** Under the 1-hop approximation and the Praos paper's RTT
+values, **the 12 MB CIP target is robustly feasible under both throughput
+models** — the Mathis-vs-CUBIC distinction is immaterial at every tested
+size.  The 1-hop assumption itself is the remaining load-bearing input;
+see §5.3 and `plots/network_model_comparison.svg`.
```
RandallYe: It would be better to make $5.3 and the diagram as hyperlinks. So people can just click it, for convenience and precision.

RandallYe: Or are you supposed to display svg image here, or just want to point to the file?

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+
+**TxCache miss rate $\pi_1$.**  The prior Haskell model fixed
+$\pi_1 = 1/6 \approx 0.167$ via a hand-chosen Markov chain.  Extracting
+$\pi_1$ empirically from the `post-cip/mempool-measurements` dataset (three
```
RandallYe: I have no idea where to find post-cip/mempool-measurements

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+(RFC 8312, the modern Linux default since 2.6.19) respond differently to
+packet loss: their steady-state throughput scales as $p^{-1/2}$ (Reno/Mathis)
+vs $p^{-3/4}$ (CUBIC), giving CUBIC ~11× more throughput at $p = 10^{-4}$.
+Because EB diffusion deadlines are tight (7 s for voters, 14 s for the full
+round), the choice of throughput model could in principle determine whether
+a 12 MB closure certifies or fails the quorum threshold — so we evaluate
+both and report whether the gap matters in the empirically-supported
+operating regime.
+
+**Key findings (corrected per-tx timing, both network models, $p = 10^{-4}$,
```
RandallYe, line 68: These findings are very interesting. If they could be presented in a structural manner by numbering them, it could be clearer. Even you can name them, and then refer to them later.

analysis/deltaq/improved-leios/report.md
crocodile-dentist: If the EB was produced but not certified, the next slot leader can't include a cert, so they produce an RB. The receiver in that case sees an RB and does no EB processing.

analysis/deltaq/improved-leios/report.md
crocodile-dentist: It renders fine for me - what do you see that is wrong with it?

analysis/deltaq/improved-leios/report.md
crocodile-dentist: It's a subdirectory of ouroboros-leios, ie. this project, https://github.com/input-output-hk/ouroboros-leios/tree/main/post-cip/mempool-measurements. Is there a better way to reference that?

RandallYe: Something like [post-cip/mempool-measurements](https://github.com/input-output-hk/ouroboros-leios/tree/main/post-cip/mempool-measurements)

or permanentlink
[post-cip/mempool-measurements](https://github.com/input-output-hk/ouroboros-leios/tree/5fd577c2ed5d5c620f57b078d84bd6562e22f4de/post-cip/mempool-measurements)

analysis/deltaq/improved-leios/report.md
RandallYe: I also found there is ! between $\Phi$ and $($ on both Firefox and Chrome. If you want a space, you could use other ways. It seems `\!` or `\;` is not rendered to space.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
+# Improved ΔQ Model for Linear Leios EB Diffusion
+
+## Table of Contents
+
+- [1. Executive Summary](#executive-summary)
```
RandallYe, line 5: These hyperlinks in toc don't work.  I think the formatted heading names inside `[# ]` are not correct.

RandallYe: Correct cross-references are very convenient for people to read through and jump around the context. These include not only those in the toc, but also links to all files, local sections, resource from other repos.

analysis/deltaq/improved-leios/report.md
RandallYe: There already have many interesting results displayed in images. It might be a good idea to render them all in this report because this is the main interface for people to read. Currently, I cannot see the image directly in this report, only tables.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+
+For $N = 12{,}288$ transactions (12 MB closure at 1 kB/tx, the CIP-0164
+target) the CPU validation time is approximately
+
+$$T_\text{CPU} \sim \mathcal{N}\!\left(N\mu_\text{eff},\; N\sigma^2_\text{eff}\right)
+= \mathcal{N}\!\left(1{,}757\,\text{ms},\;(40\,\text{ms})^2\right)$$
+
+with 99th percentile ≈ 1.85 s (Normal CLT).
+
+*Caveat on $\sigma$.*  The values $\sigma_\text{apply} = 0.527$ and
```
RandallYe: `$\sigma$` inside `* *` is not rendered correctly. Could be `*Caveat on* $\sigma$. `

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+size under both throughput models, so $P_\text{cert}$ stays at the Praos
+cap (§5.1).  See `timing_derivation.md` §7 for the CPU-only derivation.
+
+### 4.5 Certification Probability
+
+$$P_\text{cert} = P_\text{quorum} \times (1 - P_\text{interrupted})$$
+
+It is assembled from three independent sub-probabilities.
+
+#### 4.5.1 $P_\text{validating}$ {#p-validating}
```
RandallYe: `{#p-validating}` still can be seen in the rendered version.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+- **CPU reapply** — mean $N \mu_\text{eff} = 12{,}288 \times 0.143$ ms
+  $\approx 1.76$ s, with 99th percentile ≈ 1.85 s.
+
+The combined pipeline gives $P_\text{validating} \approx 0.948$ under
+Mathis and $\approx 0.992$ under CUBIC at 12 MB — both well above the
+0.75 quorum threshold.  $P_\text{validating}$ falls only slightly with
+closure size (see §5.1), so the quorum threshold is met with substantial
+margin under both throughput models for every $S_{EB-tx} \leq 12$ MB.
+
+#### 4.5.2 $P_\text{quorum}$ {#p-quorum}
```
RandallYe: similarly `{#p-quorum}`

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+
+Under **CUBIC**, the same fetches complete faster: $P_\text{validating}$
+is ≈ 1.000 at small sizes and 0.992 at 12 MB.  $\mu_V = 0.992 \times 600
+= 595$ — essentially at the upper limit — so $P_\text{quorum} \approx
+1.000$.
+
+In both models $P_\text{cert}$ stays at the Praos cap of $\approx 0.497$
+for every CIP-target closure size; see §5.1 for the full per-model tables.
+
+#### 4.5.3 $P_\text{interrupted}$ {#p-interrupted}
```
RandallYe: {#p-interrupted}

analysis/deltaq/improved-leios/report.md
RandallYe: There is another problem for the hard-coded reference like `$5.3$. If this report is to be updated in the future, all these must be manually updated. This is error-prone.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+2. **Average transaction size = 1 kB.** The CIP uses 2 kB in some places. At
+   2 kB/tx, a 12 MB closure has half as many tx hashes (so a smaller EB
+   body) and half the missing-closure-fetch size for cache misses.  Since
+   12 MB is already feasible at 1 kB/tx, the 2 kB/tx case
+   would only widen the feasibility margin.
+
+3. **Single scalar $\pi_1$ rather than a Markov chain fit.**  §5.5 and
+   `pi1_derivation.md` now provide an empirically-derived steady-state
+   value ($\pi_1 \approx 0.06$ cross-region BAU mean, 0.085 worst-pair),
+   replacing the prior model's hand-chosen $p = 0.5,\, q = 0.9$ Markov
```
RandallYe, line 1439: similar to `\!` before, `\,` is also not rendered as a space.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+   recoverable.
+
+5. **No adversarial behaviour.** The model assumes honest nodes that always
+   diffuse transactions promptly via tx-submission before the EB is produced.
+   An adversary that deliberately withholds transactions from tx-submission
+   invalidates the 1-hop approximation and could force the full multi-hop
+   diffusion scenario, reducing the feasible range significantly.
+
+6. **TCP throughput model assumptions.**
+   The transfer-time model uses TCP slow start from $\text{cwnd}_0 = 10$ MSS
```
RandallYe: `$\text{cwnd}_0 = 10$` is not rendered correctly.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1507 @@
...
+cap, no cliff.  The chart illustrates that the model corrections change
+the interpretation fundamentally: what previously appeared to be a
+feasibility limit at 0.974 is now revealed as the Praos 50% cap.
+
+**Counterintuitive appearance — blue below orange.** The $S_{EB-tx}=0$ curve
+(blue) sits *below* the $S_{EB-tx}=12\,\text{MB}$ curve (orange) at early
+times, which looks backwards at first glance.  Both curves show the
+*combined* round outcome:
+
+$$\text{cdf\_comb}(t) = p_\text{cert} \times \text{certRB}(t) + (1-p_\text{cert}) \times \text{txRB}(t)$$
```
RandallYe: There is an error displayed for this formula. 
`'_' allowed only in math mode`. 

And the formula is not rendered.

RandallYe (approved):
This report adds very interesting findings to the prior report, particularly the analysis based on the comparison of two TCP congestion-control models. This report is well-structured and the results have good data support.

My comments are all suggestions.

analysis/deltaq/improved-leios/report.md
crocodile-dentist: I see now, I was rendering html with pandoc.

analysis/deltaq/improved-leios/report.md
crocodile-dentist: I'm not sure why, they do work for html rendered with pandoc.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1547 @@
...
+   and 99% under CUBIC.
+
+7. **No FFD (Freshest First Delivery).** Following the prior report, FFD is
+   not modelled; in practice it can delay older EBs in favour of newer ones.
+
+8. **The voting deadline computation uses `3·L_hdr + L_vote = 7s`.** This
+   implies three pipeline stages each requiring L_hdr = 1 slot before the voter
+   can start. If the actual voter deadline is shorter (e.g., `L_hdr + L_vote =
+   5s`), the feasible S_EB_tx is correspondingly smaller.
+
```
bwbush, line 1519: I think it would be adding that the formulation assumes normality of some of the underlying distributions but these may in actuality have long tails. We really don't have sufficient telemetry to judge this, however, though we do have limited evidence that measurements of `apply` and `reapply` have long tails.

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1547 @@
...
+   corrected values are transaction-weighted means from mainnet blocks with at
+   most 385 transactions.  Extrapolating to EB closures of thousands of
+   transactions assumes the same per-tx cost distribution, which may not hold
+   if closure composition differs from mainnet blocks.
+
+---
+
+## 9. Plot Guide
+
+Plots are organised into three groups:
```
bwbush, line 1671: Could we inline these plots into the markdown?

analysis/deltaq/improved-leios/report.md
```diff
@@ -0,0 +1,1547 @@
...
+Mathis fails, CUBIC succeeds on the network delivery alone — but the
+voter pipeline plus quorum threshold still makes the full-blended case
+effectively infeasible under both models (see [§5.3 "Full-blended worst
+case"](#full-blended-worst-case-catastrophic-pre-diffusion-failure)).
+
+---
+
+## 10. Artifact Index
+
+| File                                       | Description                                |
```
bwbush, line 1945: Hyperlinks, please.

bwbush (approved):
This is a great addition and improvement. I read the markdown documents and examined the plots, but did not review the code itself.

analysis/deltaq/improved-leios/report.md
yveshauser: But in that case the RB can propose as well a new EB.
An EB needs to be fetched regardless of whether it is certified.

analysis/deltaq/improved-leios/report.md
crocodile-dentist: Fixed.

analysis/deltaq/improved-leios/report.md
crocodile-dentist: I've added a note here highlighting this limitation. I've also expanded section 4.4 which now shows a table that highlights the empirical visibility into the heavy tailedness of the distribution so at least it's palpable.
I've added note 9 to recommendations section, which specs out an intermediate step we could take with existing data: fit residuals from the §3.5 multivariate regression (timing_report.md) to a heavy-tailed distribution. That would give a Plutus-step-controlled per-tx variance estimate with explicit tail parameters, which is a step above our variance bounding from 4.4.

analysis/deltaq/improved-leios/report.md
crocodile-dentist: Please re-review section 3.1 to see that the current approach should be conservative, and taking your approach doesn't affect the feasibility results.

yveshauser: `p_cert` is not a-priori known, it is a value that we get out of this model. 

The analysis provides the `compute_p_certified` function to compute `p_cert` that takes the model as optional argument `voter_cdf_fn`. Shouldn't this be mandatory?

crocodile-dentist: `p_cert` is computed by `compute_p_certified` like you're saying and then used by `cdf_validate_rb` to weight the certRB vs txRB path on the receiver. Therefore `p_cert` is not some arbitrary/apriori value that the user picks and it is in fact something we do get out of this model.
Looking at `compute_p_certified`:
```python
  def compute_p_certified(s_eb_tx_kb, ..., voter_cdf_fn=None):
      ...
      _voter_fn = voter_cdf_fn if voter_cdf_fn is not None else cdf_validate_eb_for_voter
```
The default is `cdf_validate_eb_for_voter` i.e., the standard voter pipeline (EB body + 1-hop missing closure + CPU reapply). The optional argument exists so the 5.3 "full-blended worst case" analysis can substitute a different voter pipeline (one that uses blended-multi-hop missing fetch).

`p_cert` is a function of `s_eb_tx_kb` (closure size, among others) and it is also varied when running sweeps across different EB closure sizes in our delta q model.

I've updated section 3.1 to make this clearer.

analysis/deltaq/improved-leios/report.md
crocodile-dentist: @bwbush Let me know if the new additions are sound and desirable.

vcgalpin:
Some late comments (which I submitted as an issue because I'm a bit of a GitHub novice). Please see #935

PR #891, opened 2026-05-13
Add LinearLeiosTuner.html

nfrisby:
An interactive calculator/plotter for ballparking the consequences of various L values (where L_diff is the assumed bulk).

Three general consequences are at-least-somewhat illustrated:

- Expected duration between EBs succeeding and also the variability in that duration.
- Expected resulting throughput capacity (no variability here) *AND* how bad the ratio of network usage could be compared to that capacity (the paragraphs under the capacity chart).
- Pull-demand on the Mempools to refill.

### DIFF
--- a/docs/targeted-design-investigations/LinearLeiosTuner/LinearLeiosTuner-README.md
+++ b/docs/targeted-design-investigations/LinearLeiosTuner/LinearLeiosTuner-README.md
@@ -16,7 +16,7 @@ The corresponding Dynamic HTML file will replace these macros with dynamic eleme
 
 ## Plot 1
 
-The first plot is titled "Arrival rate of inter-RB gaps ≥ L (at 5% active slot coefficient)".
+The first plot is titled "Arrival rate of inter-RB gaps ≥ L on a g-chain".
 It's a step plot portraying the probability mass holding times in a renewal process defined by L (truncating x-axis after 95% cumulative).
 Repeatedly toss a coin that's biased 19:1 towards heads.
 The renewal process jumps when a run of L or more heads ends.
@@ -52,7 +52,7 @@ The stems for 25%, 50%, and 75% are taller than the others, and 50% taller still
 The second plot is titled Theoretical average capacity (with 100% stake participating, max EBs, succeed if L gap).
 Its output is another step plot, showing MaxSize divided by the mean holding time for all run lengths from 0 to ⌈L×4÷3⌉, with the value for L highlighted.
 It has a second y-axis, which is the same shape as the first y-axis, but each y2 label is the same-height y1 label÷MaxSize×100% per second.
-There is also a constant line plotted in orange at y2 = 5% per second; its describe as the protocol's worst-case protocol network usage in order to achieve the corresponding capacity.
+There is also a constant line plotted in orange at y2 = f = 5% per second (f, not g!); its describe as the protocol's worst-case protocol network usage in order to achieve the corresponding capacity.
 
 ## Template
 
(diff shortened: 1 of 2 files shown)


nfrisby:
Rendered:

- The tool https://raw.githack.com/input-output-hk/ouroboros-leios/nfrisby/ldiff-tuner-html/docs/targeted-design-investigations/LinearLeiosTuner/LinearLeiosTuner.html
- The Markdown spec file that I bootstrapped Claude from https://github.com/input-output-hk/ouroboros-leios/blob/nfrisby/ldiff-tuner-html/docs/targeted-design-investigations/LinearLeiosTuner/LinearLeiosTuner-README.md

ch1bo:
The probability distribution of possible certifications being a cut-off binomial makes sense, I guess?

I don't understand the max capacity calculation though.

Shouldn't it be just a multiplication with the expected value for P_certified? i.e. something like 48.8% for L=14

How does this assessment compare against the one used in the cost estimate and cost estimator tool
- https://leios.cardano-scaling.org/cost-estimator/
- https://github.com/input-output-hk/ouroboros-leios/pull/872

Note that the cost estimator is flipped in it's parameterization and takes a resulting throughput (= your capacity?)

docs/targeted-design-investigations/LinearLeiosTuner/LinearLeiosTuner.html
```diff
@@ -0,0 +1,366 @@
...
+    <input id="rungap-n" type="number" min="2" max="200" step="1" value="14">
+    <input id="rungap-n-slider" type="range" min="2" max="200" step="1" value="14" aria-label="L slider">
+  </div>
+  <div id="rungap-plot" class="plot"></div>
+  <div id="rungap-readout" class="readout"></div>
+</section>
+
+<section class="tuner-block" data-block="throughput">
+  <h2>Theoretical average capacity (100% stake participating, every EB full &amp; successful if L gap)</h2>
+  <p>For each run length k from 2 to &lceil;L&times;4&divide;3&rceil;, plots MaxSize divided by the mean holding time of the renewal process defined by k (which is MaxSize &middot; q &middot; p<sup>k</sup>). The step at k = L is highlighted.</p>
```
nfrisby: In https://github.com/input-output-hk/ouroboros-leios/pull/891#issuecomment-4447862369 @ch1bo said

> Shouldn't [the capacity] be just a multiplication with the expected value for P_certified? i.e. something like 48.8% for L=14

The probability of some EB being certified on-chain (assuming the cert itself exists on time) is different than the probability of two subsequent on-chain certs being X slots apart. (Surely related, but I haven't yet worked that out---might be simple.)

The first section calculates the distribution of "slots between on-chain certificates" (given a lot of optimistic assumptions). The second calculation simply divides the maximum EB size by the mean of that distribution (and the mean of the same distribution for small values of L, for visual context/reference.)

nfrisby: Yeah, I think it's pretty direct relationship, after poking it for a while.

The mean holding time is 1 / (q * p^L). The corresponding capacity estimate is MaxSz * q * p^L (ie each holding time delivers MaxSz on chain). But q is 1/20 and p^L is 48.8% for L=14. So that expression is also equal to 48.8% * MaxSz / 20, which is the arithmetic we've used before.

So this calculator (and its derivation) merely focuses on the mean holding time interpretation (ie mean time between EB successes) instead of the 48% success per attempt interpretation.

nfrisby: Hmm. Weird. In the past, the 48.8% * MaxSz / 20, I've thought of the 20 as "average time between EB's being announced". But in this calculator derivation, the 1/20 (ie q) is there because that's the chance of the cert-containing RB arising. I suppose that discrepancy is nothing more than the difference between left-closed and right-closed intervals 🤷, since the cert-containing RB and the EB-announcing RB both have the same possibility of existing (ie q=0.05).

docs/targeted-design-investigations/LinearLeiosTuner/LinearLeiosTuner-README.md
```diff
@@ -0,0 +1,41 @@
+LeiosTuner.html is a interactive Dynamic HTML calculator with two inputs.
```
nfrisby: I'm going to add a third plot: `MaxSize * q`, ie the data rate we'd see if every RB announced a full and _non-overlapping_ EB.

nfrisby: Err, I instead just added an annotation to the 2nd plot that represents this nicely.

Note the new "efficiency" readout; that's the real "pain" of increasing L and then increasing MaxSize to compensate for the slower arrival rate. The throughput capacity went down when L did, but the Leios network utilization _did not_ necessarily also go down proportional (b/c at least the adversary can still be sending non-overlapping EBs).

docs/targeted-design-investigations/LinearLeiosTuner/LinearLeiosTuner.html
nfrisby: I reorganized the page into a "reactive" document; now the text spells out this relationship very directly.

docs/targeted-design-investigations/LinearLeiosTuner/LinearLeiosTuner-README.md
```diff
@@ -0,0 +1,146 @@
...
+For a given positive maximum EB size {input MaxSize, default 12000}, we can calculate the throughput capacity offered by the successful EBs on an average g-chain.
+If we optimistically assume that every EB that g doesn't prevent is both successful and full, then the average capacity will be MaxSize ÷ (N ÷ g) = {eval MaxSize ÷ (N ÷ g)} bytes per second.
+Here it's shown for possible values of L from 1 to ⌈L×4÷3⌉.
+
+{insert plot 2}
+
+Also shown in orange is the average rate at which EBs are announced.
+Note that as L increases or g decreases, the capacity decreases but the amount of announcements doesn't change---it's still one per 1/f=20 seconds.
+MaxSize can be increased to bring the capacity back up, but that also proportionally increases the size of each announced EB.
+It's therefore important to acknowledge that adversarial RBs are able to announce any EB they want, and the whole network will diffuse it---they can choose to always announce full EBs, even if no EB has succeeded for a long while.
```
nfrisby, line 112: In https://github.com/input-output-hk/ouroboros-leios/pull/891#issuecomment-4447862369, @ch1bo asked

> How does this assessment compare against the one used in the [...] cost estimator tool?

This paragraph (and the subsequent paragraphs is one key departure. It considers a potential behaviors that the cost estimator did not. The cost estimator said:

> In the asymptotic steady state we assume tx diffusion reaches all nodes before any EB is certified, so no separate "missing tx fetch" traffic is needed — the normal tx diffusion already covers EB closure.

But the text in this Linear Leios Tuner considers two other possibilities:

- overt adversarial behavior of issuing new txs in EBs even without EBs succeeding on the chain
- similarly "non-overlapping" EB traffic as a result of mempool fragmentation

nfrisby: I got that cost estimator quote from https://github.com/input-output-hk/ouroboros-leios/blob/main/docs/cost-estimate/04-egress.md

PR #894, opened 2026-05-19
simple TCP model for sim-rs

coot:
- **sim-rs: add TCP congestion-window model**
- **sim-rs: implement TcpConnection (Chan.TCP port)**
- **sim-rs: wire tcp-congestion-control param to TcpConnection**

According to `claude`:

  | Feature | Haskell | Rust |
  |---|---|---|
  | TCP congestion-window model | ✅ | ✅ |
  | Mini-protocol multiplexing | ✅ | ❌ (`multiplex-mini-protocols` is Haskell-only) |
  | Fair bandwidth sharing across mini-protocols | N/A (mux handles it) | ✅ |

I think we need multiplexing to faithfully model TCP traffic on nodes.

### DIFF
--- a/data/simulation/config.schema.json
+++ b/data/simulation/config.schema.json
@@ -607,8 +607,9 @@
       "type": "boolean"
     },
     "tcp-congestion-control": {
-      "description": "Only supported by Haskell simulation.",
-      "type": "boolean"
+      "description": "Use the TCP congestion-window model for all links (slow-start, bandwidth-delay product, ACK pipelining). Supported by both Haskell and Rust simulations.",
+      "type": "boolean",
+      "default": true
     },
     "shard-count": {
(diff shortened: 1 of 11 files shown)


sandtreader:
@coot looks good to me - are you happy for this to be merged (it's still draft ATM).  I'd like to do it soon before it gets buried in a bunch of other stuff :-)

coot:
Yes, it can be merged.

coot:
@sandtreader one thing that we might want to change is that it is on by default, this will make more sense when we add a muxer.

I didnt know how to fix the CI issues.

sandtreader:
@coot
> @sandtreader one thing that we might want to change is that it is on by default, this will make more sense when we add a muxer.

Yes good point - will make off by default
 
> I didnt know how to fix the CI issues.

Various bits of unrelated CI in this repo are broken.  The sim-rs test is fine, that's all we need to worry about!

PR #895, opened 2026-05-20
 shared-consensus: extract 'pure' consensus core, adopt in sim-rs and net-rs

sandtreader:
## Summary

- Lifts the Praos + Linear-Leios consensus state machines out of `net-node` into a new `shared-consensus` crate under `shared-rs/` (originally extracted as `con-rs`). Sans-IO — no `tokio`, no networking, no clock reads — `Instant` and randomness are parameters; effects are returned as `Vec<Effect>`. `net-rs` and `sim-rs` both consume it via `path = "../../shared-rs/consensus"`.
- Adapts `sim-rs` to the new crate via a `SharedConsensus` variant (`leios-variant: shared-consensus`) so both consumers share the same chain selection, EB voting, vote aggregation, pipeline math, mempool, and committee selection.
- Adds a behaviour subsystem (`shared-consensus/src/behaviour/`) so adversarial / experimental nodes plug in via reactive / decision / strategy / notification hooks. Ships `LazyVoter` (policy abstention, defaulting to a new `Declined` `NoVoteReason`) and `RbHeaderEquivocator` (N-way RB-header equivocation with per-peer outbound transform).
- Net-rs fixes surfaced during smoke testing on top of the extracted crate:
  - push-on-admit `ProvideTxs` fanout in `net-node/src/main.rs` so TxSubmission diffusion doesn't stall after the startup flurry (`MissingTX` went from dominating NoVotes to essentially zero on the 30-node sample).
  - `[production.committee_selection] type = "StakeCentile"` as the default for smoke clusters in `mainnet.toml` — the wFA-LS reserve assumes mainnet-scale pool counts, which the 9-pool smoke cluster can't provide.
  - `sample-cluster.toml` carries a non-zero `tx_rate` by default so the smoke cluster actually exercises EB production / voting / cert end-to-end.

## What's in the diff (top-level)

| top-level | files |
|---|---|
| `shared-rs/consensus` | new crate (extracted from net-node) |
| `net-rs/`             | adopts the crate; mempool + telemetry plumbing; fixes |
| `sim-rs/`             | new `SharedConsensus` variant + adapter |
| `data/`               | sample topologies / parameter updates |

## Verification

- `cargo build --release` succeeds in both `net-rs/` and `shared-rs/`.
- All 316 net-core + 119 net-node tests pass on the rebased tip.
- All 270 shared-consensus tests pass.
- Smoke-cluster lazy-voter sweep on `sample-cluster-lazy-voter.toml`, identical 5-min runtime and topology per row:

| fraction | lazy / honest | cert rate |
|---|---|---|
| 0.0  | 0 / 9 | 92% |
| 0.1  | 1 / 8 | 92% |
| 0.2  | 2 / 7 | 92% |
| 0.25 | 3 / 6 | 92% (at threshold) |
| 0.34 | 4 / 5 |  0% (below threshold) |

## Test plan

- [ ] CI green on the rebased branch
- [ ] `cargo test` clean in `net-rs/` and `shared-rs/`
- [ ] Smoke clusters (`sample-cluster*.toml`) certify end-to-end
- [ ] Sim-rs `SharedConsensus` variant produces a sweep CSV matching the Linear-Leios reference numbers


buddhisthead:
Wow, it's even too big for co-pilot! Maybe you can steer us to the important bits?

sandtreader:
Additional work on UI - shows EBs being announced and certified in the chain tree.

Merging since it's all new work other than the new mode in the sim.

PR #912, opened 2026-05-27
net-node: stable CPU + memory under cluster load

sandtreader:
## Summary

Four focused fixes for the test-cluster's CPU runaway and overnight stability, identified during a 25-node smoke session. Each lives on its own commit so reviewers can take them piecemeal.

| Commit | What | Why |
|---|---|---|
| `acfb807bf` | shared-consensus: O(log P) idle peek via per-peer "unannounced" set | `peek_unannounced_for_peer` was scanning the entire mempool on every TxSubmission poll — `O(N · log A · 32)` per call. Added an inverse-partition `peer_unannounced: BTreeMap<PeerId, BTreeSet<TxId>>` populated by admit-fanout and drained by peek. Empty polls now return in `O(log P)` without touching `txs`. |
| `e27fdf991` | net-core: BlockFetch demuxer cap uses `INGRESS_LIMIT` | `set_ingress_limit` enforces a *buffer* cap on the per-protocol ingress queue, not a per-message cap. The server can pipeline `MsgStartBatch` + `MsgBlock` in a single TCP read; per-state caps of 65 KB (`StBusy`) or even 2.5 MB (`StStreaming`) race against the codec's state transition and reject legitimate pipelined block bodies. Overnight a post-EB-overflow RB body hit 2,506,268 bytes and froze the cluster. Switched to `INGRESS_LIMIT = 230 MB` (the spec's buffer cap) for both receive states. Per-message rejection at `SIZE_LIMIT_*` belongs in the codec at decode time (not yet wired). |
| `fb1fbd3e0` | net-core: front-prune LeiosStore notifications alongside slot eviction | `LeiosStore::notifications: Vec` grew monotonically forever even though the blocks/votes those entries referenced were slot-window-evicted. Switched to `VecDeque` + `notifications_pruned_count` so caller cursors stay monotonic across pruning; front-prune entries whose every referenced slot is below the cutoff. `notifications_after` now takes `&mut usize`. |
| `dca80e7a4` | net-node, shared-consensus: demote per-tx / per-event spam to debug | Three `info!` lines (`transaction received`, `network event`, `mempool: evicting oldest tx`) fire once per item under steady load and together account for ~90% of log volume at `RUST_LOG=info`. At 25 nodes × 1 tx/s/node × 12h that's enough to fill a disk. Demoted to `debug!`; periodic state-size summaries stay at `info!`. |

## Test plan

- [x] All 75 affected unit tests pass (`shared-consensus mempool`, `net-core blockfetch`, `net-core leios_store`)
- [x] 25-node release-build smoke test (~7 min): 0 errors / 0 warns, 9 RBs produced + diffused + validated across the cluster, 100% Leios certification rate, log volume ~7 MB vs ~50 MB on the old verbose build at the same elapsed time
- [ ] Longer-soak run to confirm the BlockFetch fix + notifications prune hold through overnight (test-combo branch has been the reproducer; CI doesn't currently cover this).

### DIFF
--- a/net-rs/net-core/src/peer/server_handlers.rs
+++ b/net-rs/net-core/src/peer/server_handlers.rs
@@ -529,7 +529,7 @@ pub async fn serve_leios_notify(ln_send: CodecSend, ln_recv: CodecRecv, store: A
 
         match msg {
             LnMsg::MsgLeiosNotificationRequestNext => {
-                let pending = store.notifications_after(read_index);
+                let pending = store.notifications_after(&mut read_index);
                 if let Some(notification) = pending.first() {
                     read_index += 1;
                     let response = match notification {
@@ -554,7 +554,7 @@ pub async fn serve_leios_notify(ln_send: CodecSend, ln_recv: CodecRecv, store: A
                         if subscription.changed().await.is_err() {
                             return;
                         }
-                        let pending = store.notifications_after(read_index);
+                        let pending = store.notifications_after(&mut read_index);
                         if let Some(notification) = pending.first() {
                             read_index += 1;
                             let response = match notification {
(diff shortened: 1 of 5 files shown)


sandtreader:
Closed - major issues with message sizing

PR #926, opened 2026-05-29
Technical debt: limiting votingPanel resource requirements

shd:
Currently voting/block information in net-rs/net-ui/../VotingPanel is not limited by constants, so it creates excessive load and unwanted visualisation effects in case of big networks. 

This PR introduces hard limits for VotingPanel size, while still trying to keep necessary information visible:
1. Counters for votes/blocks added: number of votes/availabe eb's is displayed in VotingPanel for each slot.
2. Number of node lines in VotingPanel is limited by 20.
3. When VotingPanel is disabled, no actual data processing for the panel is done (no requests to server, etc).


sandtreader (approved):
All looks sensible, subject to Copilot's comment about vote counts getting out of sync.

PR #929, opened 2026-06-02
Mempool events for displaying txs load in Throughput dashboard

bladyjoker:
DONE
- [x] Use Mempool traces for displaying Tx Load (removed tx-centrifuge logs based display)
- [x] Set All tooltip
- [x] Fix units and simplify expressions

image

### DIFF
--- a/demo/proto-devnet/config/dashboards/proto-devnet-throughput.json
+++ b/demo/proto-devnet/config/dashboards/proto-devnet-throughput.json
@@ -127,7 +127,7 @@
         },
         "tooltip": {
           "hideZeros": false,
-          "mode": "single",
+          "mode": "multi",
           "sort": "none"
         }
       },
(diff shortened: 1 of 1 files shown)


demo/proto-devnet/config/dashboards/proto-devnet-throughput.json
```diff
@@ -250,8 +208,8 @@
           },
           "direction": "backward",
           "editorMode": "code",
-          "expr": "rate({service=\"tx-centrifuge\", ns=\"TxCentrifuge.Builder.NewTx\"}[30s])",
-          "legendFormat": "tx/s",
+          "expr": "rate({type=\"cardano-node\", ns=~\"Mempool.AddedTx\"}[30s])",
```
ch1bo, line 211: This is good!

But this is another metric than the centrifuge releasing txs. Can we have both?

For example if we ingest only at node1 with centrifuge, their rates will always correspond. But when the centrifuge does a round robin we should see the policy effective, especially when one node can't keep up with what was ingested.

ch1bo (approved):

demo/proto-devnet/config/dashboards/proto-devnet-throughput.json
bladyjoker: You're right! Let me add it

bladyjoker:
An update version of this is in https://github.com/input-output-hk/ouroboros-leios/pull/992

PR #930, opened 2026-06-03
Node config and script for the Leios testnet

ch1bo:
Can be used with or without nix and should help getting a node targeting the Leios testnet easily.

Contains a similar setup as the `demo/proto-devnet`. That is, a `process-compose` setup of
- a single `cardano-node` configured to connect to the testnet
- a `cardano-cli` polling the tip and mempool info
- the x-ray observability stack (loki, prometheus, grafana)

See README for instructions and ways to configure.

image


bladyjoker (approved):
Brilliant! Love it!

PR #940, opened 2026-06-10
Stub out sim-rs/net-rs/shared-rs after move to leios-tools

buddhisthead:
## What

`sim-rs/`, `net-rs/` and `shared-rs/` have been extracted into their own repository, [cardano-scaling/leios-tools](https://github.com/cardano-scaling/leios-tools), with full git history preserved. This removes their source from the monorepo and leaves a short stub README in each pointing at the new home.

## Changes

- **Emptied the three directories**, each replaced by a `README.md` "has moved" notice linking to `https://github.com/cardano-scaling/leios-tools/tree/main/<dir>`.
- **Removed orphaned CI** that built the deleted source:
  - `.github/workflows/sim-rs.yaml`
  - `.github/workflows/conformance-linear.yaml` (built sim-rs + ran linear conformance against `sim-rs/parameters`)
  - `.github/workflows/net-node-docker.yaml` (net-node docker image; docker CI now lives in leios-tools)
- **Dockerfile**: dropped the Rust build stage (`rs-builder`) and the `rs` target image that compiled `sim-cli` from `sim-rs/`. The Haskell `hs` image is unchanged. No CI workflow builds the root Dockerfile's `rs` target, so nothing in CI depends on this.

Net: 339 files changed, ~196k lines removed.

## Deliberately left alone

Stale prose/comment references that don't affect the build remain and can be cleaned up separately if desired:
- `.dockerignore` (`sim-rs/target/`, `sim-rs/output/`, `!sim-rs/`)
- `topology-viewer/scripts/build-viewer-data.py` (comment mentioning `sim-rs/scripts/...`)
- comments in `data/simulation/config.default.yaml` and `leios-trace-verifier/conformance-traces/config.yaml`

> [!NOTE]
> If the root `ouroboros-leios/sim-rs` Docker image is built by any out-of-CI release process, that process should be retired or pointed at leios-tools.

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -26,20 +26,22 @@ nix run .#ui-live
 
 ## Add a scenario from sim-rs
 
-To prepare a scenario to visualize, find or add the topology to the public directory, for example:
+The Rust Leios simulator (`sim-rs`) and a few baseline topologies (`small.yaml`,
+`thousand.yaml`) now live in [cardano-scaling/leios-tools](https://github.com/cardano-scaling/leios-tools).
+Bundled topologies in `public/topologies/` are kept here directly.
 
-```sh
-mkdir -p public/topologies
-ln -sr ../sim-rs/test_data/small.yaml public/topologies/small.yaml
-```
-
-And generate a trace to visualize using the built `sim-rs`, for example using the CIP scenario:
+To add a new topology, drop the YAML file in `public/topologies/`. Then generate
+a trace to visualize with a build of `sim-rs` from the leios-tools repo, for
+example using the CIP scenario:
 
 ```bash
 mkdir -p public/traces
 cat ../analysis/sims/cip/experiments/NA,0.200/config.yaml \
   | jq '."tx-start-time" = 20' > public/traces/config-200txkbs.json
-../sim-rs/target/release/sim-cli -p public/traces/config-200txkbs.json public/topologies/small.yaml public/traces/small-200txkbs.jsonl -s 120
+/path/to/leios-tools/sim-rs/target/release/sim-cli \
+  -p public/traces/config-200txkbs.json \
+  public/topologies/small.yaml \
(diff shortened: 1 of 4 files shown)


sim-rs/test_data/small.yaml
ch1bo, line 1: Must: fix the build of `ui/`

`example.yaml`, `small.yaml` and `thousand.yaml` were referred to by the `ui/` package. See also: https://github.com/input-output-hk/ouroboros-leios/pull/940/checks?check_run_id=80669086906. Vendor them there.

Dockerfile
ch1bo, line 1: Could: just remove the whole file (along with `.dockerignore`)

I don't think this image is built/used by anymore. @will-break-it @bwbush @SupernaviX what's your memory / gut feeling about just dropping this?

ch1bo (changes requested):
This looks mostly complete. See the comments and also:

- Must: Update the top-level README to sound less stale https://github.com/input-output-hk/ouroboros-leios/blob/0ca3cf2752de358807299211a47964ce5b094ebf/README.md#L19-L21
- Could: Point directly where it lives now instead of indirection through the sim-rs/README.md here https://github.com/input-output-hk/ouroboros-leios/blob/0ca3cf2752de358807299211a47964ce5b094ebf/ui/README.md#L32-L33 mentions `sim-rs`.

sim-rs/test_data/small.yaml
sandtreader: Done as suggested

Dockerfile
sandtreader: Awaiting confirmation but agree it looks redundant now.

sandtreader (approved):
Looks good, number of lines is scary!  I've fixed the small.yaml (etc.) issue.  Approved subject to tidying the README and (maybe) deleting Dockerfile.

ch1bo (approved):
Saw the other PR with the README changes now, approved.

PR #942, opened 2026-06-17
demo/proto-devnet: don't start tx-centrifuge until nodes are ready

nfrisby:
When I ran `run.sh` locally, `tx-centrifuge` would immediately fail, complaining about the Node1 socket file not being present.

This is Claude's suggested fix, which seems reasonable to me.


bladyjoker (approved):
Nice

PR #956, opened 2026-06-25
Leios development status report

### LINKED ISSUE #909: Status report and updated leios-design
## Why

Having a progress report to update stakeholders on the Intersect Milestone 6 of the current Leios proposal.

## What

- A document that outlines the current status and links into the leios-design and other documents where suitable
- Ready by June 23rd

ch1bo:
Collected notable development progress, design challenges and highlights across all notable components / workflows of Leios. While this is in support of an intersect milestone, it should also give anyone working on or following Leios development a good overview of what happened.

But what am I explaining.. go read it :)

[Rendered](https://github.com/input-output-hk/ouroboros-leios/blob/status-report/docs/status-report-2026q2.md)

Note: `leios-design` only extended with bullet points of what should be updated too and the status report is already anticipating the digest of our GDoc, slack, etc knowledge into it.

### DIFF
--- a/docs/leios-design/README.md
+++ b/docs/leios-design/README.md
@@ -535,6 +535,8 @@ Furthermore, the existing `forgeBlock` method and/or the `BlockForging` interfac
 >
 > FIXME: also write about including certificates in blocks when available, also link to certification and subsequent sections on chain selection/block validation
 
+The first version of the Mempool can be naive, with the block production thread handling everything. A second version can try to pre-compute in order to avoid delays (ie discarding the certified EB's chunk of transactions) when issuing a CertRB and its announced EB.
+
 ### Endorser block diffusion
 
 > [!WARNING]
@@ -555,7 +557,7 @@ Furthermore, the existing `forgeBlock` method and/or the `BlockForging` interfac
 
 To satisfy **REQ-PrioritizeFreshOverStaleLeios** (freshest-first delivery), the Consensus layer must implement fetch decision logic (**NEW-LeiosFetchDecisionLogic**) that prioritizes younger EBs over older ones. This logic determines which EBs to request from which peers.
 
-Even the first version of LeiosFetch decision logic should consider EBs that are certified on peers' ChainSync candidates as available for request, as if that peer had sent both MsgLeiosBlockOffer and MsgLeiosBlockTxsOffer. A MsgRollForward implies the peer has selected the block, and the peer couldn't do that for a CertRB if it didn't already have its closure.
+Even the first version of LeiosFetch decision logic should consider EBs that are certified on peers' ChainSync candidates as available for request, as if that peer had sent both MsgLeiosBlockOffer and MsgLeiosBlockTxsOffer. A MsgRollForward implies the peer has selected the block, and the peer couldn't do that for a CertRB if it didn't already have its closure. (TODO: link to chain selection where this is relied upon)
 
 > [!WARNING]
 >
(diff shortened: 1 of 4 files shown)


docs/leios-design/README.md
```diff
@@ -462,22 +462,74 @@ Contention for the following primary node resources might unacceptably degrade t
...
+>
+> TODO: anything generic to add on top of what is in the CIP? if not -> no point having this section and instead describe the network protocols (and how they make sense for our node) in the respective consensus sections
 
 ## Consensus
 
 CIP-0164 implies functional requirements for the node to issue EBs alongside RBs, vote for EBs according to the rules from the CIP, include certificates when enough votes are seen, diffuse EBs and votes through the network layer, and retain EB closures indefinitely when certified. The Consensus layer is responsible for driving these operations and coordinating with the Network layer (which implements the actual mini-protocols) to ensure proper diffusion.
+    
+### High-throughput mempool
+
+> [!WARNING]
```
bladyjoker, line 517: Mempool is a component in `cardano-node` that "stages" transactions submitted by users of the network that haven't yet been "committed" on-chain (i.e. haven't been included in a forged Block).

Functionally, the Mempool aims to achieve three primary goals:

1. **Accept** new transactions into the network from "local clients" (e.g. user wallets via LocalClient),
2. **Diffuse** (propagate) "valid" transactions across the network (e.g. peers via TxSubmission),
3. **Pre-compute** transaction validation and state change (e.g. timely inclusion in a block during block production).

The reality, however, is that as of today, goals 2 and 3 aren't fully being met. While we could get away with these inefficiencies under Ouroboros Praos traffic loads, they become a massive bottleneck for the throughput targets under Ouroboros Leios.

To explain the challenges and opportunities, it's useful to view the Mempool as both: 

1. **Concurrent state** that can be affected through **adding and removing transactions** and **rebasing to some ledger state**. These actions can occur simultaneously and the existing behaviour is roughly just serializing these actions via a wait lock (mutex). Mempool's capacity is drastically increased under Leios and as a result rebasing now takes much longer which in conjunction with high transaction load increases wait times significantly.

2. **Task scheduler** that is managing transaction **validation** and their **application** to one or more ledger states. This distinction is crucial in the Mempool and leveraged in the existing implementation, albeit not to its fullest potential.  **Transaction validation** (through `applyTx`) has to **only ever be performed once, can be cached and performed independently from state** (with some caveats), which is great because that is the most expensive part. **Transaction application** (through `reapplyTx`) however is inherently sequential (also with some caveats) and has to be performed on each (base) state changes.

The distinction between the two modes, **validation** and **application**, can be depicted using jigsaw puzzles. Validating a puzzle piece means checking its features are aligned with the overall puzzle (brand, color, shape, thickness, material), applying a puzzle piece means finding whether it fits somewhere in the currently assembled state.

Bringing it back to transactions...

Validating a transactions means checking signatures, scripts, sizes, values etc. Applying a transaction however, means checking whether the state it builds upon to produce new state exists (utxos, stake pool, reward accounts, proposals etc).

To hammer this down it's useful to say that a transaction can be **valid** and **applicable**. If a transaction is "valid" it can (but doesn't have to) be "applicable". On the other hand, if a transaction is "applicable" then it can (but doesn't have to) be "valid". We usually want both, but it's useful to know their relationship as it will help us define what we want to diffuse (valid? valid and applicable?) and how to organize tasks in the Mempool for optimal performance.
 
Q: What about UTxOs during validation, don't you need to first fetch them from storage, isn't that necessarily state dependent? Doesn't that dissolve the distinction with "application"?
A: Yes and no! 
  - Yes, in that UTxOs form part of the state in Cardano that needs to be **resolved** before we can even start validating transactions. In other words, transactions in Cardano are **incomplete** and need a separate step before validation that could makes sense to name **transaction resolution**. How we resolve transaction UTxOs today is by looking it up in the storage associated with the currently selected ledger state. For example, the Mempool will try and resolve a transaction against its current base ledger state, if it fails the entire transaction is considered invalid. However, that is not necessarily the only or the best strategy. Perhaps the transaction can be resolved with a different base ledger state and it might become **resolveable** in a few moments when the selection changes.
  - No, in that if a transaction validates it's valid forever, even though state it refers to might have been changed and made it **unapplicable**.

ch1bo: Good stuff! Please write and push directly onto this or `leios-design-updates` branch :heart_hands:

pagio (approved):
Looks good my side

RandallYe (approved):
Read the "Praos-over-Leios prioritization" section and find it is informative and helpful.

docs/leios-design/README.md
sandtreader: I like the jigsaw analogy!

docs/leios-design/README.md
```diff
@@ -462,22 +462,74 @@ Contention for the following primary node resources might unacceptably degrade t
...
+>
+> FIXME: write up problem statement: reapplyTx on thousands of txs taking too long to block tx submission (link to other section?) and also recomputing the mempool snapshot in/for block forging is insufficient too.
+> FIXME: Write about two concrete consequences:
+>     - decouple tx diffusion from mempool syncing -> need to re-apply txs added while syncing
+>     - only block producers: keep a view (or two for Leios in Dijkstra) of txs to put into blocks -> only do slot dependent checks in ledger?
+> TODO: More advanced DAG-style mempool model?
```
sandtreader, line 523: I considered this for parallelism in Acropolis, but I'm not sure how many inter-tx dependencies exist in reality within a given transient mempool?

Issue #961, opened 2026-06-29
Extend the trace translator to support EBs


javierdiaz72:
### Why

The development of the Linear Leios node has been completed and currently there is a public testnet for it. We would like to be able to analyze the node's log files to check whether the node implementation conforms to the Linear Leios formal specification.

### What

Currently, only the Praos-related events in the log files are checked against the specification. We need to be able to also check Linear Leios-related events, in particular EB generation and reception.

### How

We will extend the trace translator to handle the `EBGenerated` and `EBReceived` trace events.

javierdiaz72:
The JSON format of the trace log file that the trace verifier expects is described [here](https://github.com/input-output-hk/ouroboros-leios/blob/main/data/simulation/trace.shared.schema.json). According to this, here's the expected format of the EB-related events:

- `EBGenerated`: Emitted whenever an EB is generated. Its format is the following:
  ```json
  {
    "type": "EBGenerated"
  , "id": "…"
  , "slot": …
  , "pipeline": null
  , "producer": "…"
  , "size_bytes": …
  , "input_blocks": null
  , "endorser_blocks": null
  }
  ```
where the value of `producer` is the node ID, the value of `slot` is the slot number in which the block was generated, the value of `id` is the block ID, the value of `size_bytes` is the total size of the block (in bytes), and the values of `pipeline`, `input_blocks` and `endorser_blocks` are `null` because they correspond to full Leios, not Linear Leios.

- `EBReceived`: Emitted whenever an EB is received. Its format is the following:
  ```json
  {
    "type": "EBReceived"
  , "id": "…"
  , "slot": …
  , "pipeline": null
  , "producer": "…"
  , "sender": "…"
  , "recipient": "…"
  }
  ```
where the value of `producer` is the node ID of the block producer, the value of `slot` is the slot number in which the block was produced, the value of `id` is the block ID, the value of `sender` is the node ID of the sender, the value of `recipient` is the node ID of the recipient, and the value of `pipeline` is `null` because it corresponds to full Leios, not Linear Leios.

The Cardano Tracer messages that correspond to the events described above are the following:

- For `EBGenerated`: The event with namespace `Consensus.LeiosKernel.TraceLeiosKernel` and kind `LeiosBlockForged` indicates that the node forged an EB. Its format is:
  ```json
  {
    "kind": "LeiosBlockForged",
    "closureSize": <???>,
    "ebSize": <size of the block in bytes>,
    "hash": <hash value of the block>,
[...]

PR #962, opened 2026-07-01
Extend the trace translator to support EBs

javierdiaz72:
This PR solves #961.


yveshauser (approved):
👍 

We could consider to base the trace verifier directly on the node log file, rather than doing the transformation to the simulator log format

PR #963, opened 2026-07-01
Leios node trace-verifier

yveshauser:
The PR introduces running the trace-verifier on the log output of a live node. The trace-verifier accepts configuration via command-line arguments and then connects to the running node in order to retrieve the stake distribution and for calculating the winning-slots of the EBs per epoch.

The trace-verifier can be invoked as in the following example:

```bash
tail -F -n +1 tmp-devnet/node1/node.log \
    | nix run '.#trace-parser:exe:linear-leios-trace-verifier-chain' -- \
        --socket-path tmp-devnet/node1/node.socket \
        --testnet-magic 164 \
        --shelley-genesis tmp-devnet/node1/shelley-genesis.json \
        --stake-pool-id pool1awyxt3egwmunup7nmd2uznqr54p2lgdt3tvrqetj8geqgfz26x9 \
        --vrf-signing-key-file tmp-devnet/node1/keys/vrf.skey
```


leios-trace-verifier/src/LinearLeiosVerifierChain.agda
```diff
@@ -0,0 +1,279 @@
...
+           ++ ((Base₂-Action s , inj₁ FFDT.SLOT) ∷ ebRole)
+           ++ vtRole
+           ++ ((Slot₁-Action s , inj₁ (FFDT.FFD-OUT (blksToHeaderAndBodyList (FFD-blks a)))) ∷ [])
+
+      traceEvent→action : Accumulator → ChainEvent → Accumulator × List Step
+      traceEvent→action a (CSlot s) =
+        if not (started a)
+          then (record a { curSlot = primWord64ToNat s ; started = true } , [])
+          else
+            let block = closeSlot a
```
javierdiaz72: It would be clearer if `block` was renamed to `blockSteps`.

yveshauser: Changed to `steps`

javierdiaz72 (approved):
LGTM

PR #973, opened 2026-07-07
Improved DeltaQ as jupyter notebook

yveshauser:
- Self-contained jupyter notebook: [analysis.ipynb](analysis/deltaq/improved-leios/analysis.ipynb)
- Results for *Full Closure Diffusion Conditional on Certification*
- PDF document of the full report (omitting the code): `nix run .#pdf`


yveshauser:
Attached the generated PDF 
[analysis.pdf](https://github.com/user-attachments/files/30186386/analysis.pdf)

yveshauser:
Regenerated PDF: [analysis.pdf](https://github.com/user-attachments/files/30220052/analysis.pdf)

vcgalpin:
I am seeing some lack of precision in talking about slot leaders and rounds that could lead to problems in the analysis. This applies to both the current document and the earlier one (https://github.com/input-output-hk/ouroboros-leios/blob/main/analysis/deltaq/improved-leios/report.md)

From the current analysis, we have the following (as one of the three distinct work streams)
> 2. Receiver RB processing of this round's RB. By protocol the RB is mutually exclusively either a certRB (probability p_cert) or a txRB (probability 1-p_cert), never both, never neither. Captured in cdf_validate_rb. Forcing  P(EB exists) here would put every RB on the certRB path, which the protocol forbids: **in failed-cert rounds the slot leader literally cannot include a cert (it does not exist) and emits a txRB instead**.

We can consider two slot leaders here.
- S0 is the slot leader that issued the RB (RB0) that is being processing by node N0 (this round) and S0 <> N0
- S1 is the slot leader that will issue the next block RB1 - this may or may not be node N0 

Taking the sentence "in failed-cert rounds the slot leader literally cannot include a cert (it does not exist) and emits a txRB instead", which slot leader does this refer to?

This cannot be S0 since it is S1 that includes the certificate for S0's EB. So the failed-cert round was the round where S0 issued the block, and not the round where S1 is issuing the block.

For S1's round, assuming certification failed on S0's round,  my understanding is that there is no requirement on S1 to issue a txRB. S1 could still issue an EB for certification and the certificate would appear in the block of the next slot leader S2 (if successful). This is independent of the certification procedure or the success/failure of this procedure for the S0's EB.

crocodile-dentist:
@vcgalpin 
Point 2 reads clear to me.

> For S1's round, assuming certification failed on S0's round, my understanding is that there is no requirement on S1 to issue a txRB. S1 could still issue an EB for certification and the certificate would appear in the block of the next slot leader S2 (if successful). This is independent of the certification procedure or the success/failure of this procedure for the S0's EB.

My understanding is that if S1 doesn't produce a txRB, it forfeits its round (acts as an adversary in an absent stake kind of way). It cannot produce an RB with a cert, because there isn't one for a round that failed (either on cadence alone, or by failed quorum), and that leaves it no other choice. The fact that it produces a txRB is separate from announcing its EB for the next round.

The txRB is just notation for an RB that contains only transactions, because in Linear as it is now, it is either or, and the notation only helps us in prose to identify which kind we're talking about, but it is one and the same object.

vcgalpin:
@crocodile-dentist wrote:
> @vcgalpin Point 2 reads clear to me.
> ...
> The fact that it produces a txRB is separate from announcing its EB for the next round.

To clarify, does this mean that a slot leader can put transactions in the RB body (assuming no certificate from the EB announced by the previous slot leader) as well as create an EB?

The text "in failed-cert rounds the [next] slot leader literally cannot include a cert (it does not exist) and emits a txRB instead" suggests that it is required to emit a (non-empty) txRB even if it is creating an EB. Is this correct?

crocodile-dentist:
> To clarify, does this mean that a slot leader can put transactions in the RB body (assuming no certificate from the EB announced by the previous slot leader) as well as create an EB?

Yes, it can produce a txRB even if it *had* a cert if it was "motivated" strongly enough to have its particular txs go on chain, but that is a separate issue. Minting RBs and announcing EBs is on the same schedule, but other than that they are separate events as far as network messages/protocol goes.

> The text "in failed-cert rounds the [next] slot leader literally cannot include a cert (it does not exist) and emits a txRB instead" suggests that it is required to emit a (non-empty) txRB even if it is creating an EB. Is this correct?

It doesn't have to produce anything on the round it won, but if it wants to announce an EB it has to announce at least a txRB. The description talks about normal non-adversarial situations. If there is no cert, then it emits a txRB, and can announce its EB for the next round.

crocodile-dentist:
@yveshauser I would like to leave my report.md in the repo and carry the burden of keeping both in sync. The markdown file is static and easy to diff for changes. I've started making some changes to my version of the report in response to new findings from my investigations, and wanted to merge your changes into my report to that end. I've noticed that your section 5.7 now references a different committee model in section 4.6:

|source|committee|threshold|
|---|---|---|
| mine 4.6.2 | 600-member Poisson sortition (each SPO elected w.p. 1−e^{−m̃·sᵢ}), has election variance (σ_V≈11) | τ·m = 450 of 600 |
| yours 5.7 | Deterministic top-stake cover — the K=923 pools covering σ_c=99% of stake, each voting with stake weight, no election variance | τ·S_active = 699 |
| 2026w18 sim (wfa-ls) | Hybrid: 480 persistent + 120 non-persistent voters = 600 | - |

The sim's `wfa-ls` splits the committee into persistent voters (480 - large pools / always in) and non-persistent ones (120 - sortition elected). It looks like yours deterministic top-stake choice may be a deliberate idealization of that persistent majority, while mine captures the sortition flavor. I realize there is still some discussion about how to establish the committee so maybe yours is what is what today's decision leans toward, but I wasn't closely following lately so I would appreciate some guidance here. Was the CIP updated here?

edit:
For me the question is whether yours deterministic committee is the more faithful idealization then I will adopt it, but if it is just a tractability device for your work's closed-form in 5.7 then I'd prefer to keep mine in 4.6.2 because there is a good reason to do so (but gloss a "different idealization" in 5.7), unless yours is the true faithful one to the CIP.

yveshauser:
@crocodile-dentist The static markdown file is easier to diff. On the other hand the jupyter notebook facilitates prototyping. But since the code is mostly generated diffing becomes more important than prototyping and I agree to keep the old setup and back-port the conditional certification diffusion sections.

yveshauser:
The voting committee selection is what currently is defined in the CIP. It was updated in the following commit https://github.com/cardano-foundation/CIPs/commit/5690adcad5701e6289c9c2532355a6c8eb77226e

yveshauser:
Closing as superseded by #1051

PR #977, opened 2026-07-08
DONOTMERGE scratch pad for Alexey to review

nfrisby:


docs/leios-design/README.md
```diff
@@ -931,6 +931,69 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+        - Is any single tx too big? (too many bytes, or too many ExUnits, etc)
+        - Is the cumulative size of all txs in the EB too big? (too many bytes, or too many ExUnits, etc)
+        - (Other components---eg ChainSync clients, LeiosFetch decision logic, etc---may have already validated the byte sizes, but it's harmless for the Ledger to double-check those while its checking the other dimensions too.)
+        - Is the EB's sequence of txs a valid extension of the ledger state of the RB that announced it?
+    - Thus, a small set of functions suffice for the needs of the LeiosVoting thread.
+        - `initializeLeiosVotingState :: LedgerState -> LeiosVotingState` - initialize this check's state from the ledger state of the announcing RB; the multidimensional tx-size accumulators start at 0.
+        - `applyTxForLeiosVoting :: LeiosVotingState -> Tx -> Maybe IsValid -> Either Error (ValidatedTx, LeiosVotingState)` - apply this transaction to the incoming state
+            - If the `Maybe IsValid` is `Just`, then Consensus has successfully validated this tx before, so the Ledger can skip state/Phase2/etc checks.
```
lehins, line 960: Extra maybe is not needed. Instead we can accept an Either. Because once a Tx has been validated you get access to ValidatedTx. It is important to do this, because we need to keep track the protocol version in which it was validated. Also there is a whole bunch of memoization we can do upon initial validation. Most importantly however, we get type safety and it would be impossible to to write code that didn't validate the Tx fully at some point
```suggestion
        - `applyTxForLeiosVoting :: LeiosVotingState -> Either Tx ValidatedTx -> Either Error (ValidatedTx, LeiosVotingState)` - apply this transaction to the incoming state
```

docs/leios-design/README.md
```diff
@@ -931,6 +931,69 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+              (Just like the Mempool's `applyTx` versus `reapplyTx` distinction.)
+            - Should return `Left` if the tx is bigger in some dimension than the current protocol parameters' values allow.
+            - Should return `Left` if the inclusion of this tx in the EB pushes its cumulative size in some dimension(s) beyond what the current protocol parameters' values allow.
+            - (If it returns `Right`, then it has increased the accumulators by the size of this tx.)
+            - Should return `Left` if the tx is not a valid extension of the accumulated ledger state.
+            - (TODO I'm assuming the correct `IsValid` flag can be projected from `ValidatedTx`)
```
lehins, line 966: As I've mentioned above, we don't need to deal with `IsValid`
```suggestion
```

docs/leios-design/README.md
```diff
@@ -931,6 +931,69 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+        - `forgetLeiosVotingState :: LeiosVotingState -> LedgerState` - discards the accumulators.
+          Leios doesn't actually _require_ this, but it seems plausible a stake pool would want to retain and reuse this ledger state if it were to later apply an CertRB that certifies this same announcement (see below).
+    - If all of the voting rules (including the Ledger checks above) are satisfied, the node will issue a vote, offering/sending it to its downstream peers.
+- When a node receives a vote from its upstream peer, it needs to validate it.
+  The vote identifies the claimed committee member by seat number and also claimed EB announcement by the announcing RB's slot and HeaderHash.
+  Because the Leios committee is stable, the node should be able to use any LedgerState.
+  Thus, we at least need `lookupLeiosCommitteeVK :: LedgerState -> SlotNo -> Int -> Maybe LeiosVertificationKey`.
+  But maybe the Ledger could also do the crypto check, so `validateLeiosVote :: LedgerState -> SlotNo -> Int -> HeaderHash -> LeiosVoteSignatureByteString -> Either Err ()`.
+  (TODO maybe EpochNo instead of SlotNo?)
```
lehins, line 975: Yes, since LeiosCommittee is stable for the whole epoch, then it is better to use EpochNo, unless EpochNo is not readily available to pass as an argument here, in which case Ledger can resolve SlotNo -> EpochNo in the rules using Globals

docs/leios-design/README.md
```diff
@@ -931,6 +931,69 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+            - (TODO I'm assuming the correct `IsValid` flag can be projected from `ValidatedTx`)
+        - `forgetLeiosVotingState :: LeiosVotingState -> LedgerState` - discards the accumulators.
+          Leios doesn't actually _require_ this, but it seems plausible a stake pool would want to retain and reuse this ledger state if it were to later apply an CertRB that certifies this same announcement (see below).
+    - If all of the voting rules (including the Ledger checks above) are satisfied, the node will issue a vote, offering/sending it to its downstream peers.
+- When a node receives a vote from its upstream peer, it needs to validate it.
+  The vote identifies the claimed committee member by seat number and also claimed EB announcement by the announcing RB's slot and HeaderHash.
+  Because the Leios committee is stable, the node should be able to use any LedgerState.
+  Thus, we at least need `lookupLeiosCommitteeVK :: LedgerState -> SlotNo -> Int -> Maybe LeiosVertificationKey`.
+  But maybe the Ledger could also do the crypto check, so `validateLeiosVote :: LedgerState -> SlotNo -> Int -> HeaderHash -> LeiosVoteSignatureByteString -> Either Err ()`.
```
lehins, line 974: It makes sense to just do the crypto check in ledger as well

docs/leios-design/README.md
```diff
@@ -931,6 +931,69 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+  (TODO maybe EpochNo instead of SlotNo?)
+  (TODO maybe we need the codomain to also permit forecast range exceptions?)
+- If enough votes are issued and diffuse well enough, then Leios might permit the next RB to include a certificate for the EB announcement.
+    - We anticipate that the Consensus Layer will want to validate this certficate when the CertRB arrives, even if the certified EB hasn't yet arrived.
+      Thus, we at least need `lookupLeiosCommitteeVKs :: LedgerState -> SlotNo -> IntSet -> Maybe LeiosVertificationKey`.
+      But maybe the Legder could also do the crypto check, so `validateLeiosCertificate :: LedgerState -> SlotNo -> IntSet -> HeaderHash -> LeiosCertficiateSignatureByteString -> Either Err ()`.
```
lehins, line 980: Ditto, makes sense to do crypto check in Ledger here too.
Few questions about this pseudo type signature:
* By `LedgerState` you mean `NewEpochState`, right? Just a clarification, since that is the highest level of the state that ledger has 😄 
* In order to go from SlotNo to EpochNo, we'll need `Globals`
* does `IntSet` mean the BitArray with indices of StakePools?
* If answer for above question is yes, then would it be more accurate  to have this type signature where this is [`LeiosCert`](https://github.com/IntersectMBO/cardano-base/blob/319385df3b1e856713686fc68a7b07698f269fc7/cardano-crypto-leios/src/Cardano/Crypto/Leios.hs#L235-L238) definition:
```suggestion
      But maybe the Ledger could also do the crypto check, so `validateLeiosCert :: Globals -> NewEpochState era -> SlotNo -> HeaderHash -> LeiosCert -> Either Err ()`.
```

* Furthermore, at this point Ledger has validated the RB that announced the EB, so I'd actually recommend storing that HeaderHash in the NewEpochState, which would make these guarantees stronger that correct header is actually used. Therefore this signature can further be reduced to:

```haskell
validateLeiosCert :: Globals -> NewEpochState era -> SlotNo -> LeiosCert -> Either Err ()
```

* Is the `SlotNo` in the argument of when RB with EbAnnouncement was applied or the SlotNo when Rb with LeiosCert will be applied? If I understand correctly, it must be the SlotNo of when the block with EbAnnouncement was applied, since upon construction of the certificate SlotNo of RB with certificate was not even known yet. This means we can also store that SlotNo in the NewEpochState, thus reducing this function to:
```haskell
validateLeiosCert :: Globals -> NewEpochState era -> LeiosCert -> Either Err ()
```

In order to avoid Ledger re-validating the cert and have the proof by the type system that it was indeed validated, we can construct a type that is opaques to consensus and return from this function:
```haskell
data LeiosCertValidity = LeiosCertValidity
  { lcvCert :: LeiosCert
  -- ^ Validated certificate
  , lcvRbHash :: HeaderHash
  -- ^ Hash of the block header that announces the Eb
  , lcvSlotNo :: SlotNo
  -- ^ SlotNo of announcement
  } deriving (Eq, Show)

validateLeiosCert :: Globals -> NewEpochState era -> LeiosCert -> Either Err LeiosCertValidity
```
This data type will be passed as signal to `applyCertifiedLeiosBlock`, which means we'll be able to just compare the type to the expected one in the `NewEpochState`, instead of doing expensive cryptographic validation once again.

docs/leios-design/README.md
```diff
@@ -931,6 +931,71 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+The Ledger engineers might want to interpret that type as `NewEpochState` (although `LedgerState` also contains other state as well).
+
+There are two high-level ways in which Leios interacts with the Ledger.
+
+- Fully applying a CertEB to a ledger state must incorporate that EB's transactions into the resulting ledger state.
+- The ledger state determines the Leios committee for each epoch.
+
+The following seems like a useful (roughly chronological) order in which to consider these.
+
+- When a block-producing node acquires an EB, the Leios rules sometimes require them to issue a vote for its EB announcement.
```
lehins, line 948: > When a block-producing node acquires an EB

Current terminology is not sufficient to describe what is meant here, namely what is meant by EB:
* a list transaction hashes with sizes
* or a list of actual transactions 

We need proper terminology for that

Hence was my suggestion on slack that defines this very nicely:
```data
newtype EbBlockHeader = EbBlockHeader
  { ebhTransactionHashes :: [(Hash Tx, Word32)]
  }

newtype EbBlockBody = EbBlockBody
  { ebbTransactions :: [Tx]
  }
```

With that definition we get:
```
data Eb =
  { ebBlockHeader :: EbBlockHeader
  , ebBlockBody :: EbBlockBody
  }
```

lehins, line 948: > When a block-producing node acquires an EB

Current terminology is not sufficient to describe what is meant here, namely what is meant by EB:
* a list transaction hashes with sizes
* or a list of actual transactions 

We need proper terminology for that

Hence my suggestion on slack that defines this very nicely:
```data
newtype EbBlockHeader = EbBlockHeader
  { ebhTransactionHashes :: [(Hash Tx, Word32)]
  }

newtype EbBlockBody = EbBlockBody
  { ebbTransactions :: [Tx]
  }
```

With that definition we get an actual block that fits into the **blockchain**:
```
data Eb =
  { ebBlockHeader :: EbBlockHeader
  , ebBlockBody :: EbBlockBody
  }
```

docs/leios-design/README.md
```diff
@@ -931,6 +931,71 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+      But maybe the Legder could also do the crypto check, so `validateLeiosCertificate :: LedgerState -> SlotNo -> IntSet -> HeaderHash -> LeiosCertficiateSignatureByteString -> Either Err ()`.
+      Note that the SlotNo and the HeaderHash are those of the announcing RB, not the CertRB that contains the certificate.
+      In particular, the HeaderHash is the CertRB's prevhash.
+      As with validating a vote, any LedgerState that's not older than the immutable tip should suffice.
+      (TODO use EpochNo instead of SlotNo?)
+      (TODO forecast exception?)
+- Once both the CertRB and the EB it certifies have fully arrived, the node might try to properly apply the CertRB to its predecessor chain.
+    - The first step would be to apply the EB to the ledger state of the announcing RB.
+      Thus, `applyCertifiedLeiosBlock :: LedgerState -> LedgerLeiosBlock -> LedgerState`.
+      The `LedgerLeiosBlock` type is whatever data the Ledger needs to do this; it must be entirely determined by the EbBody and the EbClosure, which is what the Leios components of the Consensus Layer have acquired from the network (or from self, if self-issued).
```
lehins, line 989: I don't quite like the name Leios block for Eb, since Ranking Block is also introduced with Leios. Other than the naming I totally agree with this 👍  
```suggestion
      The `Eb` type is whatever data the Ledger needs to do this; it must be entirely determined by the EbBody and the EbClosure, which is what the Leios components of the Consensus Layer have acquired from the network (or from self, if self-issued).
```

docs/leios-design/README.md
```diff
@@ -931,6 +931,71 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+      For example, maybe `LedgerLeiosBlock blk = [(GenTx blk, IsValid)]`.
```
lehins, line 990: I decided to avoid this confusing discussion that spanned from IsValid being separate from the transaction and we'll just absorb it into the transaction itself. Key part is that serialization of Tx in block will be different from Tx in mempool. See: https://github.com/IntersectMBO/cardano-ledger/issues/5912

docs/leios-design/README.md
```diff
@@ -931,6 +931,71 @@ For the first version of the LedgerDB, it need not explicitly store EB's ledger
...
+    - The contract for `applyCertifiedEb` should require that Consensus has already validated the cert (as discussed above) before calling `applyCertifiedEb`.
+      The valid cert cannot exist without at least one honest stake pool having issued a vote (as discussed above), which means the EB is known to be entirely valid against this exact same ledger state (it's the one that that pool passed to `initializeLeiosVotingState`) and therefore `appliedCertifiedEb` doesn't need to do _any_ checks at all: it merely needs to ("re")construct the resulting ledger state.
```
lehins, line 992: This totally relies on Consensus living up to the promise and executing `validateLeiosCert` with the right arguments prior to this call. I strongly believe we either need to follow Sebastian's suggestion of re-validating LeiosCert here once again, or better yet, as I've mentioned in my other comment, use Haskell's type system to keep Consensus honest and accept `LeiosCertValidity` here for validation as a proof.

docs/leios-design/README.md
lehins, line 948: > When a block-producing node acquires an EB

Current terminology is not sufficient to describe what is meant here, namely what is meant by EB:
* a list transaction hashes with sizes
* or a list of actual transactions 

We need proper terminology for that

Hence was my suggestion on slack that defines this very nicely:
```data
newtype EbBlockHeader = EbBlockHeader
  { ebhTransactionHashes :: [(Hash Tx, Word32)]
  }

newtype EbBlockBody = EbBlockBody
  { ebbTransactions :: [Tx]
  }
```

With that definition we get:
```
data Eb =
  { ebBlockHeader :: EbBlockHeader
  , ebBlockBody :: EbBlockBody
  }
```

lehins (commented):
My feedback so far

docs/leios-design/README.md
nfrisby: We cannot afford to keep all of the "recent" txs that we have validated in-memory. It could be several gigabytes in order to span the amount of time the TxCache is intended to remember things for (which means the TxCache might contains _millions_ of txs). But we _can_ afford to retain a single in-memory bit per tx.

... I suppose we could use that bit in order to _decode_ the tx directly into a `ValidatedTx`. Does that sound OK to you? Then we get both: your tighter interface and TxCache keeps its txs' bytes on disk.

docs/leios-design/README.md
nfrisby: - Yep; the scratch pad starts with:

    > As a Consensus engineer, I'm writing `LedgerState` below, thinking of the data type that the Consensus layer uses.
     > The Ledger engineers might want to interpret that type as `NewEpochState` (although `LedgerState` also contains other state as well).

- Yep, `IntSet -> ... -> LeiosCertficiateSignatureByteString ->...` was intended to be isomorphic to `LeiosCert -> ...` (which I hadn't seen in code before)

- Yes, the `SlotNo` is the slot of the announcing RB, not of the certifying RB.

- We have found in our prototype that we also need to keep the "the predecessor RB's announcement, if any" in the protocol state (ie its slot, its RbHash, and its EbHash). So certainly seems plausible to keep that in the LedgerState too. The only reason not too is because Consensus is already carrying that in the "extended" `ExtLedgerState`, so we can just provide it in the calls that need it. So: prefer a slightly wider interface or prefer duplicated state?

- `LeiosCertValidity` sounds good to me; it captures the contract that Consensus has validated the cert before calling `applyCertifiedLeiosBlock`---except for the fact that only code review and testing is preventing us from passing the wrong `LeiosCertValidity` to `applyCertifiedLeiosBlock`. If we have a bug and do pass you the wrong `LeiosCertValidity` then I'll have to think about what Consensus should do (crash? stop validating the block but _without_ interpreting it as invalid? :thinking:)

docs/leios-design/README.md
nfrisby: Looks like your next comment supersedes this one, so I'll click "Resolve conversation".

nfrisby: > When a block-producing node acquires an EB (both its body and its closure)

That's ^^^ what I had intended by "EB".

Your `EbBlockHeader` type is `EbBody`. And your `EbBlockBody` is `EbClosure`.

~Except: we'll need to add the `IsValid` flags into one of them. (Either in `EbBody` or in `EbClosure` in a way that influences the corresponding hashes and sizes in `EbBody`.)~ Edit: Ah ha, now I've seen your comment below about https://github.com/IntersectMBO/cardano-ledger/issues/5912. I'm very happy to see it.

> With that definition we get an actual block that fits into the *blockchain*:

We should synchronously discuss the motivation for emphasizing _blockchain_.

Edit: a note to self, in case that discussion doesn't happen soon. In my intuition, Leios has _two_ "header-body" divisions/layers. The EbAnnouncement and the EbBody form one header-body pair, and the EbBody and the EbClosure form another. From the protocol-level design, it seems Sebastian, Giorgos, me, etc all ended up (preferred?) prioritizing the EbAnnouncement-EbBody pairing instead of the EbBody-EbClosure pairing. So, main point is: there are _three_ objects we need to consider, not just two. The EbAnnouncement has many responsibilities that we generally associate with RbHeaders (no coincidence they're intentionally conflated in the current version of CIP-0164).

nfrisby: This comment is also redundant (it's the 3rd copy, and 2nd one seemed best), so I'm clicking "Resolve conversation" here too.

docs/leios-design/README.md
nfrisby: Agreed. Ledger would still need to confirm that Consensus has passed a `LeiosCertValidity` _for the right EB announcement_, but once you've checked that, Ledger isn't trusting Consensus about the cert at all.

After that adjustment, the interface in the scratch pad seems sufficient for everything you've mentioned except for one thing: checking the hash of the EbBody and that Consensus decoded it correctly, and so on. We could pass you the bytes we decoded the EbBody from, eg? So you can hash those. Is that preferable? And would the Tx decoder used to decode the EbClosure retain the MemoBytes for those, so you could check that those have the hashes and sizes listed in the EbBody?

PR #997, opened 2026-07-19
Mempool design

### LINKED ISSUE #911: Prototype high-throughput mempool
## Why

Both https://github.com/input-output-hk/ouroboros-leios/issues/789 and https://github.com/input-output-hk/ouroboros-leios/issues/845 identified the mempool as a bottleneck to attain high data throughput

The goal is to attempt the low hanging fruits and see whether they would be "good enough" for the initial throughput targets (for the overall consensus). e.g. for 50 TxkB/s (10x of mainnet), a 100 TxkB/s mempool should be sufficient?

## What

- Make sure we can measure time and space around the mempool
- Establish a baseline
  - A reproducible scenario that shows the throughput limit of the mempool for a given load on a given machine
  - Excluding network latencies
  - Including disk access latencies (UTxO-HD)
  - No need for high fidelity or exact reproducibility, 10% std deviation is fine
- Attempt two improvements:
  - Time- and size-bound revalidation work happening in the Forge loop
  - Being more lax in what txs are served (while synchronizing)
- Update the `leios-design` document with explored design options

## How

- [x] https://github.com/input-output-hk/ouroboros-leios/issues/971
- [x] Configure proto-devnet to run with utxo-hd and many small txs -> good baseline?
  - [ ] Make the txs use more (full limit?) exunits -> much different?
- [x] Don't starve tx diffusion while "syncing the mempool"
  - Not use a `TMVar`?
  - Make sure to re-apply any txs that got added while syncing the mempool
  - [x] https://github.com/IntersectMBO/ouroboros-consensus/pull/2229
  - [x] https://github.com/IntersectMBO/ouroboros-consensus/pull/2148
- [x] Attempt forge loop improvements and see if they improve the mempool throughput (less cross talk)
  - [x] https://github.com/IntersectMBO/ouroboros-consensus/pull/2095
    - Not sure this makes sense given 2217
  - [x] https://github.com/IntersectMBO/ouroboros-consensus/pull/2094
  - [x] https://github.com/IntersectMBO/ouroboros-consensus/pull/2217
  - [x] https://github.com/IntersectMBO/ouroboros-consensus/pull/2183
- [x] Write-up leios-design
  - [x] https://github.com/input-output-hk/ouroboros-leios/pull/997
  - [x] https://github.com/input-output-hk/ouroboros-leios/pull/999

bladyjoker:
See #911 

Initial brain dump on the Mempool topic. I think it's a good conversation starter and contains ideas about tweaks we could do today to help correctness and performance, but also hints at the ideal that we should aim for in order to effectively battle adversarial nodes.

I still have some topics to discuss, especially about EB extended ledger state treatment in the Mempool. I didn't go into DAG Mempool much, I think we're a bit far from that, however, I think it should become a more desirable topic in the near future.

Would love to hear your thoughts on this!

### DIFF
--- a/docs/leios-design/README.md
+++ b/docs/leios-design/README.md
@@ -591,7 +591,7 @@ The main questions we need to answer:
 1. How do we decide which transactions to accept and diffuse?
 2. How do we decide which transactions are eligible to be included in the new block?
 
-##### Mempool `slot`and `Tick`
+##### `Tick`
 
 `slot` is a state variable that changes on `Tick` action and can affect both transaction "validity" and "applicability".
 
@@ -654,7 +654,7 @@ In practice, this interval would be ~20 slots long, arguably an acceptable trade
 Why would that work?
 Ticking changes state observable to transactions and therefore can affect their validity and applicability only **on epoch boundary**.
 In other words, all ticking in between epoch bounds is not observable to transactions and can't affect their validity and applicability.
-So, ticking that stays within the bounds of the current Mempool transaction interval costs very little!
+So, ticking that stays within the bounds of the current Mempool transaction interval costs very little! Not just that, we don't really have to tick, if we do, it's just to help the internal "ledger" accounting perform smaller incremental computations.
 
 Let's update the model!
 
@@ -663,7 +663,7 @@ Let's update the model!
 -- the `txs` that were resolved, validated and
 -- applicable to `base` ledger state AND `interval`.
 data MempoolState = MempoolState {
-  interval :: (SlotNo, SlotNo),
+  interval :: (SlotNo, SlotNo), -- technically can be derived from `base`
   base :: LedgerState,
   txs : [Tx]
 }
@@ -680,9 +680,61 @@ data MempoolState = MempoolState {
 And back to the main questions:
 
 1. How do we decide which transactions to accept and diffuse?
-  - Transactions that are valid and applicable in `interval` are accepted and diffused.
+  - Transactions that are valid and applicable to `interval` are accepted and diffused.
 2. How do we decide which transactions are eligible to be included in the new block?
   - Transactions that are valid and applicable to `interval` which contains the election winning slot can be included verbatim in the new block.
+
+##### `Rebase`
+
+`base` is a state variable that should reflect the currently selected ledger state. When a selection changes, Mempool needs to rebase `txs` on top of this new `base` (and `interval`).
+
+Since there can be many transaction in `txs` it can take a long time to re-apply this sequence.
+It's noteworthy to reiterate that we only need to **apply** and not **validate** `txs` as we established before that **validation** can be cached forever (or for-era) while **application** is state dependent.
+
+(TODO: Insert some data here)
+
+(The very fact that we treat them as a sequence is unfortunate, as especially in Cardano ledger there's many opportunities to schedule this work more intelligently by treating transactions as a DAG by utilizing transaction topology. However, that's currently a research and development question that requires a non-trivial amount of effort.)
+
+1. on `Rebase newState`
+  1. Update `base` to `newState`
+  2. Update `interval` to `(max (tipSlot newState) previousEpochSlot, min nextExpectedBlockSlot nextEpochSlot)`
+  3. re-apply `txs`
+
+Scenario: Bob is sending Alice 100 txs per second, and Alice is going through a rebase, what happens?
+
+In the current implementation, we're utilizing a wait-lock, during a rebase all access to the Mempool is blocked until the rebase is done.
+If a rebase takes 1 second, Bob will queue up 100 transactions waiting on the rebase to finish, and in that time Bob's transactions aren't diffused.
+In a naive, lock-free solution, Bob might be able to add transactions to Alice's Mempool while the rebase is ongoing and diffusion would continue,
+but when the rebase is done and its work needs to be committed, we're faced with a similar problem:
+there's 100 transactions more in the Mempool than when the rebase started. What do we do with them? Drop them? Or re-apply them too?
+If we re-apply them, do we do that in some priority way? Perhaps with a lock?
+
+Sounds like this is an improvement! We're not waiting on a longer rebase but instead waiting on a shorter "merge" that reconciliates
+conflicts due to Mempool effects occurring concurrently (namely re-applying 100 surplus transactions).
+
+```haskell
+
+mergeMempoolState :: MempoolState -> MempoolState -> MempoolState
+mergeMempoolState lm rm  = -- in the case of miss matching `base`, 'select' the one with the "better" base and
+                           -- if there's any surplus transactions in "worse" mempool apply them in "better".
+                           ...
+```
+
+Let's take a step back at this point an go back to our main questions:
+
+1. How do we decide which transactions to accept and diffuse?
+  - Transactions that are valid and applicable to `base` are accepted and diffused.
+2. How do we decide which transactions are eligible to be included in the new block?
+  - Transactions that are valid and applicable to `base` can be included verbatim in the new block.
+
+Similar to the ticking scenario, since Bob and Alice don't know which `base` they see, they are guessing, and that can be abused by adversaries.
+
+##### `AddTx`
+
+`txs` is a state variable that holds a sequence of transactions received from either local clients or TxSubmission peers. Upon receival, every transaction first needs to be **resolved** against the in-memory cache or on-disk storage, **validated** and finally **applied** to `base` and `interval`.
+
+These 3 stages of transaction processing should be separately considered and measured to understand their processing bounds. What's the "worst" case scenario for resolution, is not the same as what it is for validation or application.
+
 
 ### Block production
 
--- a/docs/leios-design/README.md
+++ b/docs/leios-design/README.md
@@ -578,7 +578,7 @@ data MempoolAction = AddTx Tx -- peers and local clients (1000 per s)
 
 -- Mempool state holds the `base` ledger state,
 -- the `txs` that were resolved, validated and
--- applied to `base` ledger state AND `slot`.
+-- applicable to `base` ledger state AND `slot`.
 data MempoolState = MempoolState {
   slot :: SlotNo,
   base :: LedgerState,
@@ -607,21 +607,6 @@ It can also affect **applicability** because the slot can be outside of transact
 
 That being said, time also affects the ledger state itself, as some state variables are "internal" to it and defined in terms of time, and not observable by transactions (for example incremental rewards accounting). In Cardano, such effects are accumulated incrementally and "published" on epoch boundaries.
 
-(TODO: Include Alexey's epoch tick measurement data)
-
-In summary, the ticking pseudocode:
-
-1. on `Tick newSlot` (newSlot > slot)
-  1. update `slot` to `newSlot`
-  2. tick `base` to `newSlot`
-    1. update internal state variables
-    2. if `newSlot` in new epoch
-      1. update state variables observable to txs
-  3. if `newSlot` in new epoch
-    1. if `newSlot` in new era
-      2. re-validate `txs`
-  4. re-apply `txs`
-
 Why do we care about ticking in the Mempool? Well, that's the same as answering our two main questions wrt time...
 
 1. How do we decide which transactions to accept and diffuse?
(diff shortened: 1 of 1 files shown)


ch1bo:
I'm trying to pick this up this week while working on #911

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+
+#### Transaction processing modes
+
+The distinction between the two modes, **validation** and **application**, can be loosely depicted using jigsaw puzzles. Validating a puzzle piece means checking its features are aligned with the overall puzzle (brand, color, shape, thickness, material), applying a puzzle piece means finding whether it fits somewhere in the currently assembled state.
+
+Bringing it back to transactions...
+
+Validating a transactions means checking signatures, scripts, limits, values etc. Applying a transaction however, means checking whether the state it builds upon to produce new state exists (utxos, stake pool, reward accounts, proposals, **time** etc).
```
lehins, line 551: This is statement is too loose in my opinion. Validation of a transaction is not just checking signatures and limits, it is also ensuring that transaction is valid against the current ledger state, eg. withdrawal is not taking out more funds from the account that is available in it or utxo being spent is actually present in the state.

That being said, if you think of transaction that is being "applicable" is the one that is "valid" against the existing ledger state, I'd suggest changing this terminology in this document, because that does not align with the terminology that we have today.

Terminology of today is:
* "static" validation is checking all of the transaction parts that doe not depend on the state
* "non-static" validation is checking all parts of the transaction that do depend on the state

I believe this is more descriptive, because we also have a mode where simply apply transaction without validation, which can be easily confused with as "applicable"

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+
+Q: What about UTxOs during validation, don't you need to first fetch them from storage, isn't that necessarily state dependent? What if you have dependent transactions? Doesn't that dissolve the distinction with "application" and makes "validation" inherently sequential?
+A: Yes and no!
+
+  - Yes, in that UTxOs form part of the state in Cardano and need to be **resolved** before we can even start validating transactions either by querying storage or preceding transaction outputs. In other words, transactions in Cardano are **incomplete** and need a separate **resolution** step before validation that makes them **complete**.
+
+  - No, in that if a transaction validates it's valid **forever** (actually for an era), even though state it refers to might have been changed and made it **inapplicable** to some particular state. How one **resolves** transactions and makes them **complete** can be sequential, but doesn't have to.
```
lehins, line 560: This is a wild overstatement.
* It is possible for a plutus script to change its outcome after an intra-era hard fork. So even a static check, can change its behavior within an era. This is just one example, but we can have more, because that is the point of hard fork: change the validity of a transaction
* Transactions naturally become valid or invalid for a specific slotNo
* Transactions can conflict or depend on each other in many different ways, which is can only be proven through the ledger state. Albeit inter-dependency can be inferred simply through looking at any two transactions.

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+1. How do we decide which transactions to accept and diffuse?
+2. How do we decide which transactions are eligible to be included in the new block?
+
+##### `Tick`
+
+`slot` is a state variable that changes on `Tick` action and can affect both transaction "validity" and "applicability".
+
+(In Cardano, time is denoted with slots, and is handled separate to `LedgerState`, not strictly necessary but it's worth mirroring that implementation choice.)
+
+It can affect **validity** because slots passing through epoch boundaries can change protocol parameters and the "ledger law". Since validity is essentially "ledger law" judgement on transactions, a transaction that was valid in one epoch is not necessarily valid in another. In other words, validity is defined per "ledger era". For example, script limits could change between eras.
```
lehins, line 600: There is also an explicit validity interval field in a Tx. So it does affect validity directly.
I guess you probably mean "applicability" hear. But that is precisely why I have a problem with this "new" terminology. Even the name of the field literally says "validityInterval"

```suggestion
It can affect **validity** because slots passing through epoch boundaries can change protocol parameters and the "ledger law". Since validity is essentially "ledger law" judgement on transactions, a transaction that was valid in one epoch is not necessarily valid in another. In other words, validity is defined per "ledger era". For example, script limits could change between protocol versions.
```

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+
+(Q: Is it the case that if a Cardano transaction is valid in era N it's valid in all future M > N eras? Given the transaction "forward translation" feature? If so, we don't have to re-validate transactions when crossing epochs/eras?)
```
lehins, line 602: No it is not.
Forward translation feature does not guarantee translation, nor does it guarantee validity.
```suggestion
```

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+
+It can also affect **applicability** because the slot can be outside of transaction interval. Another way applicability can change is, again, when passing through epoch boundaries as certain actions are put in effect late (pool registration etc.) and some state variables are created or deleted (stake pools). For example, a transaction with "always" interval that retires a pool A can be applicable in one epoch but not another.
+
+(In the current terminology it would make sense to say transaction applicability interval, rather than validity interval)
```
lehins, line 606: As you well spotted it yourself 😄 
So, instead of changing terminology after half a decade of developers and users around the World using it, it would be better to stick to the same terminology as we have today: static vs non-static validity.
Granted, what we have today is not a commonly known terminology, so we could change that, but changing the notion of transaction "validity" as we know it today is unacceptable. 
```suggestion
```

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+1. How do we decide which transactions to accept and diffuse?
+  - Transactions that are valid and applicable to `slot` are accepted and diffused.
+2. How do we decide which transactions are eligible to be included in the new block?
+  - Transactions that are valid and applicable to `slot` which matches the election winning slot can be included verbatim in the new block.
+
+Scenario: Bob is sending a transaction X to Alice, but they don't know what `slot` the others' Mempool has.
+
+How will Alice determine to accept or reject the transaction X (and potentially punish Bob)?
+After all, there's no guarantee both Mempool `slot`s match!
+Perhaps in Bob's Mempool transaction X is valid and applicable, just on a different `slot`.
```
lehins, line 621: It is not only the slot, but also the set of transactions that Alice has can determine whether to accept Bobs transaction, there are no guarantees that they will have the same set.
Even the order in which they received transaction could affect the set of transactions they both will have in their mempool

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+  2. tick `base` to `newSlot`
+    1. update internal state variables
+    2. if `newSlot` in new epoch
+      1. update state variables observable to `txs`
+  3. if `newSlot` in new epoch
+    1. if `newSlot` in new era
+      2. re-validate `txs`
+  4. re-apply `txs`
+
+Are we really going to do this every second?
```
lehins, line 640: Ticks today happen on every slot that has a block, not every second.

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+
+Step `1.2` can be fast, but on epoch boundary it's actually quite slow.
+Step `1.3.2` is always expensive, but re-validation only has to happen on new era.
```
lehins, line 643:
```suggestion
Step `1.3.1.2` is always expensive, but re-validation only has to happen on new protocol version.
```

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+Step `1.4` can be made fast by using "smarter" scheduling techniques, but for now is also slow.
+
+Is there something better we can do?
+
+I propose the **Mempool transaction interval** that rejects transactions with interval that don't contain it.
+This intervals' left bound is the maximum between tip slot and last epoch boundary, and the right bound is minimum between "expected next block slot** and next epoch boundary.
+In practice, this interval would be ~20 slots long, arguably an acceptable tradeoff that gives correctness and performance but takes away transactions with an interval smaller than 20 slots.
+
+(TODO: Include Kosta's mainnet interval analysis)
```
lehins, line 652: 😄 
```suggestion
(TODO: Include Kostas' mainnet interval analysis)
```

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+there's 100 transactions more in the Mempool than when the rebase started. What do we do with them? Drop them? Or re-apply them too?
+If we re-apply them, do we do that in some priority way? Perhaps with a lock?
+
+Sounds like this is an improvement! We're not waiting on a longer rebase but instead waiting on a shorter "merge" that reconciliates
+conflicts due to Mempool effects occurring concurrently (namely re-applying 100 surplus transactions).
+
+```haskell
+
+mergeMempoolState :: MempoolState -> MempoolState -> MempoolState
+mergeMempoolState lm rm  = -- in the case of miss matching `base`, 'select' the one with the "better" base and
```
lehins, line 718: It is theoretically impossible to merge two separate ledger states, unless there share a common base and effects of transactions are known and transactions are disjoint in their effects.
For example, imagine there is a base ledger state with account `X` that has 10ADA in it. Then `lr` has a transactions that deposits 5 ADA into account `X`, while `rm` has a transaction that withdraws 10ADA and unregisters account `X`. Clearly `rm` is in conflict with `lr`

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+
+It can affect **validity** because slots passing through epoch boundaries can change protocol parameters and the "ledger law". Since validity is essentially "ledger law" judgement on transactions, a transaction that was valid in one epoch is not necessarily valid in another. In other words, validity is defined per "ledger era". For example, script limits could change between eras.
+
+(Q: Is it the case that if a Cardano transaction is valid in era N it's valid in all future M > N eras? Given the transaction "forward translation" feature? If so, we don't have to re-validate transactions when crossing epochs/eras?)
+
+It can also affect **applicability** because the slot can be outside of transaction interval. Another way applicability can change is, again, when passing through epoch boundaries as certain actions are put in effect late (pool registration etc.) and some state variables are created or deleted (stake pools). For example, a transaction with "always" interval that retires a pool A can be applicable in one epoch but not another.
```
sandtreader, line 604: Also protocol parameters can change on epoch boundaries, so (for example) a deposit for registration which was valid in one epoch may not be in another.  This puts protocol parameters into the "applicability" context, I think.

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+##### `Rebase`
+
+`base` is a state variable that should reflect the currently selected ledger state. When a selection changes, Mempool needs to rebase `txs` on top of this new `base` (and `interval`).
+
+Since there can be many transaction in `txs` it can take a long time to re-apply this sequence.
+It's noteworthy to reiterate that we only need to **apply** and not **validate** `txs` as we established before that **validation** can be cached forever (or for-era) while **application** is state dependent.
+
+(TODO: Insert some data here)
+
+(The very fact that we treat them as a sequence is unfortunate, as especially in Cardano ledger there's many opportunities to schedule this work more intelligently by treating transactions as a DAG by utilizing transaction topology. However, that's currently a research and development question that requires a non-trivial amount of effort.)
```
sandtreader, line 696: I looked at topological dependency ordering in Acropolis and found for the dominant case of UTXOs (there are other dependencies - e.g. registrations), because the validation (in your terms, independent of order) can be done in advance and in parallel, by the time you come to do the order-dependent stuff you may as well keep it linear, particularly as the state you're modifying (the UTXO set) has to be mutexed anyway.

docs/leios-design/README.md
```diff
@@ -524,6 +524,273 @@ CIP-0164 implies functional requirements for the node to issue EBs alongside RBs
...
+
+data MempoolAction = AddTx Tx -- peers and local clients (1000 per s)
+  | Rebase LedgerState -- chain selection (several per 20s)
+  | Tick SlotNo -- time passing (1 per s)
+
+-- Mempool state holds the `base` ledger state,
+-- the `txs` that were resolved, validated and
+-- applicable to `base` ledger state AND `interval`.
+data MempoolState = MempoolState {
+  interval :: (SlotNo, SlotNo),
```
lehins, line 766: What happens with a transaction that has a validity interval that has only a partial intersection with this interval? The only correct thing to do would be to exclude such transaction, but that would mean this would artificially reduce number of transactions that would actually make it on chain.

lehins (changes requested):
I have more questions than answers with these proposed design changes to the mempool.

First and foremost, terminology is inconsistent with what we have today.

There is some new terminology that has a very hand wavy definition. For example, what does it mean to "rebase" a ledger state and most importantly how does transaction conflict resolution happens upon such "rebase"?

It seems like this whole PR is based on an assumption that there is a trivial implementation of a `mergeMempoolState` like function.

ch1bo:
Superseded by https://github.com/input-output-hk/ouroboros-leios/pull/999

PR #999, opened 2026-07-24
Mempool design

### LINKED ISSUE #911: Prototype high-throughput mempool

ch1bo:
Alternative/extension of the basic structure started in #997 with the concrete design of separating syncing from serving.

This got now more elaborate and also drafts sections for the validity range checking and maybe keeping two read views for the forge loop?


docs/leios-design/README.md
```diff
@@ -513,41 +514,123 @@ The node must include new mini-protocols (**NEW-LeiosMiniProtocols**) to diffuse
...
+
+![](./mempool-component-diagram.svg)
+
+#### Double-buffering
+
+A transaction is fully validated only once, when it is first admitted (`applyTx`), and thereafter only re-applied against a changed base (`reapplyTx`), which is far cheaper because it reuses that validation and only re-checks that the transaction still applies (see [Transaction validation levels](#transaction-validation-levels)).
+
+This asymmetry is currently used by the mempool, and the double-buffering design leans on it further: a change of base forces `O(|txs|)` re-applications, but each one is cheap, and in particular cheaper than the full validation that concurrent additions are paying at the same time. It also means only that first validation should need to load data from an on-disk ledger; a transaction already in the mempool should be re-applied without reading from disk at all.
```
nfrisby: > a transaction already in the mempool should be re-applied without reading from disk at all.

That's not true: when the selection changes, it might have consumed UTxO that some tx in the Mempool consumed. So now that tx no longer applies. The only way to detect that is to use the updated UTxO map, which for UTxO HD may require reading from disk (ie via the new selection's `Forker`).

nfrisby: I believe my objection above also applies to the TODO just below

docs/leios-design/README.md
```diff
@@ -513,41 +514,123 @@ The node must include new mini-protocols (**NEW-LeiosMiniProtocols**) to diffuse
...
+  , txs  :: [Tx]         -- resolved, validated, applicable transactions
+  }
+
+data MempoolAction
+  = AddTx  Tx           -- validate and apply a new tx (~1000/s)
+  | Rebase LedgerState  -- re-apply all txs onto a new base (a few per 20s)
+  | Remove [TxId]       -- force-drop txs to recover from forging a rejected block (rare)
+```
+
+Reading dominates and must never block: diffusion continuously serves `txs` to peers, and block production draws from them to fill blocks. Against that steady stream of reads, the two mutations are comparatively rare. `AddTx` validates and applies a peer's or client's transaction against `base` (`applyTx`) and appends it. `Rebase` takes a new `base` from chain selection and re-applies (`reapplyTx`) every transaction onto it. The invariant both preserve is that a transaction is in `txs` exactly when it is valid and applicable to `base`, which is also the Mempool's core question: which transactions to accept, diffuse, and allow into a block.
```
nfrisby: > Against that steady stream of reads, the two mutations are comparatively rare.

That smells like an overstatement to me, but I'm not sure. There is some kind of parity: adding txs to our mempool (a "mutation") is what causes us to send those contents to our peers (a  "read").

But maybe I'm missing something? Does the TxSub consumer "read" the Mempool more than it actually sends new information to peers?

nfrisby: > `AddTx` validates and applies a peer's or client's transaction against `base` (`applyTx`) and appends it.

No it doesn't. It applies applies a new tx to the ledger state that results from applying all of the txs already in the Mempool, in order, to `base`. (If `base` is instead that "accumulated" ledger state, then `base` is misnamed.)

nfrisby (changes requested):
I'm short-circuiting my review for two reason:

- I already called out some significant inaccuracies in the explanation of the Mempool.
- I don't think this document is the correct place to be explaining how the Mempool works/what it does. We just need to explain what it does _differently_, right?
- You and I have our 1:1 in five minutes :D

docs/leios-design/README.md
```diff
@@ -52,19 +53,19 @@ The Linear Leios protocol operates by allowing a second, bigger type of block to
...
 
-Despite being hard deadlines, these targets reflect the reality that network vulnerability increases when not being met. The protocol's safety and liveness guarantees depend on honest nodes being able to propagate blocks rapidly enough to prevent adversarial forks from gaining traction. Failure to meet these timing constraints can lead to increased rates of short forks, reduced chain quality, and - if persistent - ultimately compromise the integrity of the ledger. 
+Despite being hard deadlines, these targets reflect the reality that network vulnerability increases when not being met. The protocol's safety and liveness guarantees depend on honest nodes being able to propagate blocks rapidly enough to prevent adversarial forks from gaining traction. Failure to meet these timing constraints can lead to increased rates of short forks, reduced chain quality, and - if persistent - ultimately compromise the integrity of the ledger.
```
crocodile-dentist, line 60: This is not a hard deadline, quite the opposite it is a soft one which gives Praos its work conserving properties. The 5s delta parameter only lives in the Praos paper's proof, it is not a protocol parameter and no producer/validator waits on the clock to guillotine it. Blowing the deadline still gives a block a chance to be adopted outright by the network, or it battles a diffusion-settled fork race. It might seem like hair-splitting, but this is absolutely critical to keep in mind.

ch1bo: I'm not changing it in this PR and I believe the sentence is capturing your sentiment? Would you want me to change something?

crocodile-dentist: s/hard deadlines/soft bounds/

PR #1000, opened 2026-07-27
Fix a few bugs in the ./ui visualizer

nfrisby:
Claude wrote all of this, I just guided it through the debugging process.

- I was frequently seeing disparities in the number of sends/recvs despite seeing perfect balance in the logs.
- Claude eventually brainstormed a promising explanation: using `/tail` to "live stream" from Loki is known to be incomplete (especially if some messages are big, which we've seen).
- So: this PR switches from `/tail` to `/query_range`.
- But that requires CORS, which Loki can't support, so we added a `caddy` proxy to compensate.

Deduping was one of the early brainstormed mitigations, but now it's necessary since the `/query_range` intentionally overlaps in order to pick up events that arrived to Loki after our previous poll.

~There was also a discovery that Alloy is sending things to Loki in duplicate---but I decided not to mess with that because it wasn't trivial.~

-----

I just expanded this PR with a couple more fixes.

- a bugfix to my first "./ui: switch Loki from /tail to /query_range to stop dropping events" commit, b42a23eb66cd9e95cf6e292ce4ab8ff95130eead---`/query_range` delivers events bundled by streams, where as `/tail` was in chrono-order. So the patched commit now merges the streams by chrono order. There are various parts of the visualizer that depend on the events being chrono-ordered, eg the `Parent` field in the `Chain` View.
- I also noticed the `Certifies EB` field was broken in the `Chain` View, but I suspect this wasn't due to my changes? Similar problem: parser had side-effects that assumed some ordering, but it _also_ assumed no-dups/idempotency, which is not the case. Alloy sents events multiple times, unannotated and also annotated. The fix here was to filter out annotated events, so we only process each even once (not for _all_ of the parsers, but for some of them).
- I also accidentally included a tiny but generic feature: draw edges as dashed lines until they're utilized for the first time (my demo of announcements uses a linear topology to have multi-hop paths, so one edge of the triangle remains dashed throughout the whole visualization in that case).

### DIFF
--- a/ui/src/components/Sim/hooks/lokiWorker.ts
+++ b/ui/src/components/Sim/hooks/lokiWorker.ts
@@ -1,9 +1,15 @@
-// Web Worker that owns the Loki tail WebSocket and runs the parser chain
-// off the main thread. The main thread's only job is to dispatch the
-// already-typed event batches into the reducer, which keeps the WebSocket
-// reader from falling behind Loki's internal tail buffer (the cause of the
-// silent `dropped_entries` losses we observed when this was all on the
-// main thread).
+// Web Worker that polls Loki's query_range HTTP API off the main thread.
+//
+// We poll query_range rather than the /tail WebSocket: /tail is best-effort
+// and silently omits entries (not counted in `dropped_entries`), whereas
+// query_range returns the complete window. A sliding window with overlap
+// tolerates Loki's ingestion latency, and the dedup below removes both the
+// re-scanned overlap and Loki's duplicate-ingested copies (the Alloy pipeline
+// stores each cardano line more than once).
+//
+// Loki serves no CORS headers, so the demo fronts it with a small CORS-adding
+// reverse proxy (see demo/extras/x-ray); the scenario's `loki` host points at
+// that proxy and the fetch below hits it directly.
 
 import { IServerMessage } from "@/components/Sim/types";
 import { parseStreamValue, resetPendingState } from "./lokiParsers";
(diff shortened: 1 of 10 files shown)


ui/src/contexts/SimContext/reducer.ts
```diff
@@ -131,7 +131,30 @@ export const reducer = (
...
-      const newEvents = [...state.events, ...action.payload];
+      // Keep `events` ordered by `time_s`. The aggregator relies on this: it
+      // stops scanning at the first event past the current time, so an
+      // out-of-order event would truncate the scan and silently drop later
+      // events from every count. The live Loki path interleaves
+      // independently-delivered per-direction streams, so batches arrive out
+      // of order. Merge-insert the (small) sorted batch into the already-sorted
+      // list in O(n + m); a full re-sort each batch would be O(n log n),
+      // untenable once a demo reaches hundreds of thousands of events.
+      const incoming = [...action.payload].sort((a, b) => a.time_s - b.time_s);
```
nfrisby, line 142: Ask Claude if there's a library call for this

ui/src/components/Sim/hooks/lokiWorker.ts
```diff
@@ -22,108 +28,183 @@ export type LokiWorkerResponse =
   | { type: "EVENTS"; events: IServerMessage[] }
   | { type: "DROPPED"; count: number };
 
+// Alloy writes each cardano-node line to Loki twice: once before it promotes
```
nfrisby, line 31: @bladyjoker confirmed this duplication manually; he's now intending to remove it

nfrisby (commented):
Drazen and I had a call

bladyjoker (approved):
Looks good!

ui/src/components/Sim/hooks/lokiWorker.ts
nfrisby: He'll circle back to remove this new filter once it's unnecessary

ui/src/contexts/SimContext/reducer.ts
nfrisby: Ah ha. I mispoke in the call with Drazen.

- The merge-insert logic below is inserting the new batch (actually, one per "stream") into `state.events`, which is _every event the visualizer has ever seen_ (recall that the UI's scrubber bar goes all the way back).
- So we're inserting into a sorted array that contains a number of events proportional to how long the visualizer has been running.
- So the sorted inserts do seem worthwhile.
- And Claude claims there's no off-the-shelf library routine to call.

PR #1001, opened 2026-07-28
Corrections for the "Proto DevNet - Throughput" dashboard.

geo2a:
This PR introduces a few corrections for the dashboard:

- Submission load Y axis should be TxkB/s
- Confirmed TX throughput Y axis should be TxB/s
- Mempool right Y axis should be TxkB/s
- Fix the "Time since last Leios certification or announcement" graph so that it shows events from all nodes and not only from node1

Here's a screenshot after these changes:
labels-fixed


ch1bo:
Did similar changes in https://github.com/input-output-hk/ouroboros-leios/pull/1006, can you check those out?

geo2a:
Thanks for merging @ch1bo. Sorry, I was away last week, will have a look at #1006 now.

PR #1006, opened 2026-07-29
New demo/dozen-devnet and mempool throughput changes

ch1bo:
Setup and tooling that came out of the mempool throughput work in #911. The analysis and its results live separately; this is what is needed to reproduce the runs.

## New: `demo/dozen-devnet`

A twelve-node devnet: 3 block producers, each with 3 private relays, and the nine relays fully meshed. A block producer only ever talks to its own relays, so every transaction and every block between two pools crosses at least two relay hops. That is the degree profile of a real SPO deployment, unlike proto-devnet's all-to-all.

- Traffic control per node, `RATE` and `DELAY`, defaulting to 50 Mbps and 10 ms one way. `TC=0` disables it.
- Up to three on-demand load generators, one per relay group with its own delegator key, so injection points can be added mid-run without restarting.
- Knobs: `TPS`, `OUTPUTS` (transaction size, via the generator's `--outputs-per-tx`), `NODE_RTS`, `TX_FIREHOSE` (which generator binary to run, since process-compose resolves from the PATH it started with).
- UTxO-HD with the `V2LSM` on-disk backend, so ledger access is as demanding as intended.
- `provenance.txt` per run, recording each binary's resolved path, version and sha256 alongside the settings. A nix-store node and a local cabal build report the same git rev, so this is the only durable way to tell runs apart.
- README with the host capacity limits found along the way, including the ones that invalidated earlier runs.

## Measurement harness

`scripts/sample.py` samples every node's Prometheus endpoint into a TSV: on-chain bytes from the ledger state at the tip, forge counters, and per-node mempool depth and N2N ingest. Phase is derived from the number of running generators, so starting a second one mid-run labels the rows that follow without bookkeeping.

`scripts/digest.py` extracts what the logs are for before they are deleted, which matters because a two-generator run writes 1.7 GB in twenty minutes and twice filled the disk mid-run. It reports the mempool add-path gap distribution per node (invisible to a 1 s scrape, so unrecoverable afterwards), per-generator submission stats, the forge loop by call stack, endorser block fill, and votes withheld by reason.

## proto-devnet

- Throughput dashboard reworked: ingest against submitted, submission in TxkB/s, per-generator breakdown, and log-derived counters no longer presented as node metrics.
- Tracing cut back substantially so the tracer stops being a load source at high transaction rates.
- Namespaces torn down when the project stops, plus `hold-namespaces.sh` and `teardown-namespaces.sh` shared with dozen-devnet.
- Do not assume `iproute2` is on the system PATH.

Also a `flake.lock` bump.


demo/proto-devnet/config/config.yaml
```diff
@@ -11,13 +11,21 @@ MempoolCapacityBytesOverride: 25000000
...
 LastKnownBlockVersion-Major: 3
 LastKnownBlockVersion-Minor: 1
 ConsensusMode: GenesisMode
 RequiresNetworkMagic: RequiresMagic
 LedgerDB:
-  Backend: V2InMemory
+  # On-disk UTxO-HD backend (LSM-tree backing store) so UTxO reads/writes hit
+  # disk — makes ledger access demanding, as wanted for mempool throughput
+  # baselining. Switch back to V2InMemory to keep the UTxO set in RAM.
+  Backend: V2LSM
```
geo2a, line 28: I'd be careful with enabling LSM for the demo. We do know that the nodes running LSM are capable of producing blocks, but I think we still recommend to use the in-memory backend on mainnet.

PR #1015, opened 2026-07-31
Musashi proto-devnet alloy + dash alignment

### LINKED ISSUE #995: Grafana consolidation Musashi/ProtoDevnet/TesnetRelay
@johnalotoski and myself want to make a push towards a more sustainable story in Leios observability.

Currently our Grafana dashboards and Alloy configurations live separately in Musashi, Testnet Relay (this repo) and ProtoDevnet.

To make this happen let's try and tackle two problems:
1. Data model consolidation, meaning which common fields exist for all `services`? Namely what does `environment` and `instance` mean etc. If our queries are to be portable, we have to assume certain fields to exist and mean a particular thing.
2. Sourcing/reusing Grafana dashboards and Alloy configurations from a single place (preferably).

What we'd like to see is something similar to what we have in ProtoDevnet Grafana instance. Separate logs, parsed and indexed for various purposes, most important being `cardano-node`, but also `call-trace` and `leios-voting`.

There's many gotchas here to get right into making sure the storage backend doesn't blow up and Grafana can perform well. The less "parsing" happens at use time  (`| json`) the better, which means more alloy configs which is something to manage.

To get this going let's start with porting the following dashboards to Musashi from ProtoDevnet:
1. Throughput - uses `cardano-log` service type
2. CallTrace - uses `call-trace` service type
3. Voting - uses `leios-voting` service type

johnalotoski:
Proto-devnet and Musashi Playground Alloy and Dashboard alignment:

* Unifies alloy modules and setup between proto-devnet and Musashi
* Proto-devnet alloy modules are imported and re-used in cardano-playground
* Unifies dashboards between proto-devnet and Musashi
* Dashboards for Leios are imported and re-used in cardano-playground
* The throughput dashboard maintains compatibility for both firehose and centrifuge use-cases

This PR was developed with assistance from Claude.


demo/proto-devnet/config/alloy.template
```diff
@@ -8,24 +30,24 @@ prometheus.scrape "cardano_node_exporter" {
 	targets = [
 		{
 			"__address__" = "${IP_NODE1}:${METRICS_PORT_NODE1}",
-			"job"         = "cardano_node_metrics",
+			"job"         = "integrations/cardano-node",
 			"instance"    = "node1",
-			"alias"       = "node1",
-			"process"     = "Node1",
+			"environment" = "leios",
+			"group"       = "proto-devnet",
```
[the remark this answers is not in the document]
johnalotoski: Fixed.

demo/proto-devnet/config/alloy.template
```diff
@@ -82,263 +119,90 @@ loki.process "extract_process_compose_logs" {
...
-			host  = "host",
-			sev   = "sev",
-			ns    = "ns",
+	// Normalized routing label. Modules select on `stream`, never on `process`,
+	// so they stay portable across environments.
+	stage.match {
+		selector = `{process=~"Node[0-9]+"}`
+		stage.static_labels {
+			values = {
+				stream = "node",
```
bladyjoker: How is `stream` different to `service`? And why not `cardano-node`?

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
```diff
@@ -0,0 +1,111 @@
...
+          name   = "name",
+          thread = "thread",
+        }
+        drop_malformed = true
+      }
+
+      stage.static_labels {
+        values = {
+          service = "cardano-node/call-trace",
+          type    = "leios-call-trace",
```
bladyjoker: We can drop `type` everywhere, `service` fully captures that notion?

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
```diff
@@ -0,0 +1,111 @@
...
+//   Input : node_enrich output — body = decoded `data`, `kind` label set.
+//   Output: same line, plus service=cardano-node/call-trace and bounded
+//           event/name/thread labels on Call lines; call counters emitted.
+//
+// The call-graph identifiers (id/child_id/parent_id/stack/argument) are left in
+// the body for query-time `| json` — `stack` is the call path (unbounded) and
+// must never be a label. Only event/name/thread (bounded) are indexed.
+// (NB: main's inline version had `parent_id = "child_id"` and labelled `stack`;
+// both are corrected here — parent_id/stack live in the body via `| json`.)
+declare "call_trace_enrich" {
```
bladyjoker: I find "enrich" to mean "purely additive", given that this is actually 
1. parsing and drop "malformed" inputs
2. adding labels

And that it's actually defining a "Alloy process" I would prefer `call_trace_process` as a function name.

demo/proto-devnet/config/alloy-modules/call_trace_process.alloy
```diff
@@ -0,0 +1,111 @@
...
+
+  export "receiver" {
+    value = loki.process.enrich.receiver
+  }
+
+  loki.process "enrich" {
+    stage.match {
+      selector = `{kind="Call"}`
+
+      stage.json {
```
bladyjoker, line 49: Before there was
```alloy
		stage.json {
			expressions = {
				event     = "event",
				argument  = "argument",
				id        = "id",
				child_id  = "child_id",
				parent_id = "parent_id",
				name      = "name",
				stack     = "stack",
				thread    = "thread",
			}
			drop_malformed = true
		}
```

this is important as this is where "parsing" happens, I'm enforcing a very specific structure. Even though I'm not setting most of them as labels, it's important to ensure certain fields exist.

demo/proto-devnet/config/alloy-modules/call_trace_process.alloy
```diff
@@ -0,0 +1,111 @@
...
+        }
+      }
+
+      stage.labels {
+        values = {
+          event  = "event",
+          name   = "name",
+          thread = "thread",
+        }
+      }
```
bladyjoker, line 70: 👍

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
```diff
@@ -0,0 +1,111 @@
+// call_trace_enrich — the Leios call-trace facet of the cardano-node stream.
+//
+// Chained after leios_voting_enrich; acts only on kind="Call". Others pass
+// through. Usually the tail of the node chain, so its `forward_to` is the
+// terminal write.
```
bladyjoker: I don't understand this comment. 

"Others pass through"? We drop all input that isn't of expected format.

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
```diff
@@ -0,0 +1,111 @@
...
+//
+//   Input : node_enrich output — body = decoded `data`, `kind` label set.
+//   Output: same line, plus service=cardano-node/call-trace and bounded
+//           event/name/thread labels on Call lines; call counters emitted.
+//
+// The call-graph identifiers (id/child_id/parent_id/stack/argument) are left in
+// the body for query-time `| json` — `stack` is the call path (unbounded) and
+// must never be a label. Only event/name/thread (bounded) are indexed.
+// (NB: main's inline version had `parent_id = "child_id"` and labelled `stack`;
+// both are corrected here — parent_id/stack live in the body via `| json`.)
```
bladyjoker: I would remove this, a bit too verbose

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
```diff
@@ -0,0 +1,111 @@
+// call_trace_enrich — the Leios call-trace facet of the cardano-node stream.
+//
+// Chained after leios_voting_enrich; acts only on kind="Call". Others pass
+// through. Usually the tail of the node chain, so its `forward_to` is the
+// terminal write.
+//
+//   Input : node_enrich output — body = decoded `data`, `kind` label set.
+//   Output: same line, plus service=cardano-node/call-trace and bounded
+//           event/name/thread labels on Call lines; call counters emitted.
```
bladyjoker: I like it, may I suggest a general documentation pattern, smtn like:

```
Input: JSON with fields event, argument, id, child_id, parent_id, name, stack, thread
Output: Valid input
Labels: static `service = cardano-node/call-trace`, event, name, thread
Metrics: started_total, ended_total, duration_total, allocations_total

Example:

{
...
  "kind" : "Call",
  "event": "Start",
  "name": "decide-leios-certifiy",
  "thread": "Forge",
  "id": "131987.5.0",
  "parent_id": "131987.5",
  "child_id": 0,
  "stack": "forge -> partition-mempool -> decide-leios-certifiy",
  "argument":2638248
...
}
{
...
  "kind" : "Call",
  "event": "End",  
  "name": "decide-leios-certifiy",
  "thread": "Forge",
  "id": "131987.5.0",
  "parent_id": "131987.5",
  "child_id": 0,
  "stack": "forge -> partition-mempool -> decide-leios-certifiy",
  "argument": 2638248,
  "allocations": 1027888,
  "duration": 0.008065857,
  "result": "19d1cad584e1ca66a6b5b77c2c1c1f638c0b98ba1f429019c80c29965d00a0e9"
...
}
```

demo/proto-devnet/config/alloy-modules/leios_voting_process.alloy
```diff
@@ -0,0 +1,84 @@
...
+    stage.match {
+      selector = `{kind=~"LeiosVoted|LeiosVoteAcquired"}`
+
+      stage.json {
+        expressions = {
+          voterId = "vote.voterId",
+        }
+        drop_malformed = true
+      }
+
```
bladyjoker, line 42: As before we want to make sure the events have a valid structure

```
		stage.json {
			expressions = {
				rbHash  = "vote.rbHash",
				voterId = "vote.voterId",
			}
			drop_malformed = true
		}
```

demo/proto-devnet/config/alloy-modules/leios_voting_enrich.alloy
```diff
@@ -0,0 +1,84 @@
...
+        expressions = {
+          voterId = "vote.voterId",
+        }
+        drop_malformed = true
+      }
+
+      stage.static_labels {
+        values = {
+          service = "cardano-node/leios-voting",
+          type    = "leios-voting",
```
bladyjoker: Not needed

demo/proto-devnet/config/alloy-modules/leios_voting_enrich.alloy
```diff
@@ -0,0 +1,84 @@
...
+// single write. Non-vote lines pass straight through to `forward_to` untouched.
+//
+//   Input : node_enrich output — body = decoded `data`, `kind` label set.
+//   Output: same line, plus service=cardano-node/leios-voting and a bounded
+//           `voterId` label on vote lines; vote counters emitted.
+//
+// `rbHash` is deliberately NOT extracted/labelled — it is ~one value per block
+// (high cardinality). It stays in the body for `| json` (vote.rbHash), which also
+// keeps the counters bounded (voterId ~3 is the only variable label they inherit).
+declare "leios_voting_enrich" {
```
bladyjoker: leios_voting_process?

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
```diff
@@ -0,0 +1,73 @@
...
+//
+// Pattern B, drop-first: anything that is not the node stream is dropped up
+// front, so the rest of the module operates only on its own lines and can never
+// contaminate a sibling stream. The trace body is parsed exactly once here; the
+// vote/call facets downstream guard on the `kind` LABEL (no re-parse to route).
+//
+// Cardinality: index only bounded fields. `host` is omitted (it duplicates the
+// base `instance` label); `rbHash`/`stack` are never labelled — they stay in the
+// body (republished `data`) for query-time `| json`.
+declare "node_enrich" {
```
bladyjoker: Let's use the full name rather and _process `cardano_node_process`?

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
```diff
@@ -0,0 +1,73 @@
...
+    stage.json {
+      expressions = {
+        at   = "at",
+        sev  = "sev",
+        ns   = "ns",
+        kind = "data.kind",
+        data = "data",
+      }
+      drop_malformed = true
+    }
```
bladyjoker: Same argument as before

```
	stage.json {
		expressions = {
			at     = "at",
			sev    = "sev",
			host   = "host",
			thread = "thread",
			ns     = "ns",
			kind   = "data.kind", // TODO(bladyjoker): ns="Consensus.LeiosPeer.Msg|Reflection.TracerConfigInfo|Version.NodeVersion" don't have `kind`, otherwise it's common
			data   = "data",
		}
		drop_malformed = true
	}
```

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
```diff
@@ -0,0 +1,73 @@
...
+      source = "at"
+      format = "RFC3339"
+    }
+
+    stage.labels {
+      values = {
+        level = "sev",
+        sev   = "sev",
+        ns    = "ns",
+        kind  = "kind",
```
bladyjoker: Please add thread, that's pretty useful

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
bladyjoker: I guess we don't care about `host` as it's already in `instance`?

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }
+
+    stage.static_labels {
+      values = {
+        service = "cardano-node",
+        type    = "cardano-node",
```
bladyjoker: Redundant `type`

demo/proto-devnet/config/alloy-modules/tx_centrifuge_enrich.alloy
```diff
@@ -0,0 +1,83 @@
...
+        sev   = "sev",
+        ns    = "ns",
+      }
+    }
+
+    stage.static_labels {
+      values = {
+        service   = "tx-centrifuge",
+        type      = "tx-centrifuge",
+        generator = "centrifuge",
```
bladyjoker: I think the service is enough?

demo/proto-devnet/config/alloy-modules/tx_centrifuge_enrich.alloy
```diff
@@ -0,0 +1,83 @@
...
+declare "tx_centrifuge_enrich" {
+  argument "forward_to" {}
+
+  export "receiver" {
+    value = loki.process.enrich.receiver
+  }
+
+  loki.process "enrich" {
+    stage.match {
+      selector = `{stream!="tx-centrifuge"}`
```
bladyjoker: Why not `ns !~"TxCentrifuge.*"`?

demo/proto-devnet/config/alloy-modules/tx_firehose_enrich.alloy
```diff
@@ -0,0 +1,110 @@
...
+declare "tx_firehose_enrich" {
+  argument "forward_to" {}
+
+  export "receiver" {
+    value = loki.process.enrich.receiver
+  }
+
+  loki.process "enrich" {
+    stage.match {
+      selector = `{stream!="tx-firehose"}`
```
bladyjoker: Why not `ns !~ "TxFirehose.*"`

demo/proto-devnet/config/alloy.template
```diff
@@ -1,3 +1,25 @@
...
+//     wired here. Each is a self-contained `declare` that expects a structure and
+//     forwards to an injected receiver — no module hardcodes its downstream.
+//   * Routing follows Pattern B (fan-out + drop-first) at the service boundary:
+//     the front-end forwards every line to each service module, and each module
+//     drops anything that is not its stream. The node sub-facets (voting,
+//     call-trace) are CHAINED after node_enrich so each node line is still
+//     written exactly once, carrying whatever the facets added.
+//
+//   front-end ─┬─(raw)────────────────────────────► loki.write.loki
+//              ├─► node_enrich ─► leios_voting ─► call_trace ─► loki.write.loki
```
bladyjoker: I don't think this is correct?

That's the pipeline I'm working with
Image

demo/proto-devnet/config/alloy.template
johnalotoski: We can re-purpose `service` for the routing fan-out as you suggest, dropping the separate `stream` label entirely. The one caveat is the raw process-compose copy: since a `loki.process` stamps its labels onto every `forward_to` target, overwriting `service` for routing on the same stage (instead of using a separate `stream` label) would clobber the raw copy's `service=process-compose`. So to keep that raw copy for all the processes with no enrichment module a small two-stage front-end can be made — stage 1 unwraps + writes the raw copy, stage 2 stamps service + fans out. Only proto-devnet needs that as the playground base journal pipeline already writes the raw copy, so it routes in one stage.

If it is decided that the general `service=process-compose` logs aren't needed in loki because only the enriched module processed logs are utilized, we can simplify the o-leios alloy front end processor back to a single stage like playground.

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
johnalotoski: `type` label is used in 2 dashboards from pre-existing commits, but it looks like it can easily be replaced with `service` label usage, freeing us to drop the `type` label as you suggest.

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
johnalotoski: No problem, will do.

demo/proto-devnet/config/alloy-modules/call_trace_process.alloy
johnalotoski: `stage.json` pulls declared fields into a temporary extracted map for subsequent stage use, but there is no structure enforcement here.  Everything, expression matched or not, still passes through.  The only thing dropped are log lines which legitimately don't json parse successfully and that is controlled with `drop_malformed = true`; if set `false` non-parseable json will still be passed through too.

So if a `stage.json` expression won't be used in a subsequent step, there is really no reason to include it as it incurs additional compute with no benefit.

For structure enforcement (beyond the basic criteria of being json parseable), that has to come from `stage.drop` or `stage.match { action = "drop" ... }`.

The expression vars removed above were dropped due to high cardinality.

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
johnalotoski: Sure, will do.

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
johnalotoski: At this module the only thing dropped is malformed JSON with `drop_malformed = true`.

Routing drops happen once up front when the fan-out forwards every line to each service module, and each module's drop-first guard keeps only its own respective service logs.

While the `tx.*` routes then go straight to the write in a single step, the cardano-node route is a chain instead (showing your enrichment -> process rename): cardano_node_process → leios_voting_process → call_trace_process → loki.write. Once past `cardano_node_process`'s drop-first guard, every line in the chain is a cardano-node line; each step enriches only its own sub-type and passes every other line through untouched to the next step. So at the terminal write, each node line is written exactly once, carrying whatever enrichment its sub-type warranted.

demo/proto-devnet/config/alloy-modules/leios_voting_process.alloy
johnalotoski: Similar response to above: `stage.json` does not enforce structure, `rbHash` is not indexed due to cardinality nor used in other subsequent stages so removed from the `stage.json` expression to avoid unnecessary compute. `rbHash` is still read time queryable which the voting dashboard leverages.

demo/proto-devnet/config/alloy-modules/leios_voting_enrich.alloy
johnalotoski: `type` label dropped, thanks!

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
johnalotoski: Same structure-enforcement point as above but note the only two dropped from your list are `host` and `thread`.

Re `host`: agreed, redundant with the preferential `instance`.
Re `thread`: the envelope thread is a numeric thread-id, high cardinality (more in another comment).

For the `kind` TODO, that's already handled as those namespaces are valid JSON, so `drop_malformed` does not drop them and the kind label just comes out empty.  Nothing special needed.

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
johnalotoski: Sure.

demo/proto-devnet/config/alloy-modules/leios_voting_enrich.alloy
johnalotoski: Let's do it :)

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
johnalotoski: `type` label dropped, thanks!

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
johnalotoski: The `thread` at the node level is the envelope top-level numeric Haskell thread-id, one per lightweight thread, effectively unbounded, so as a label on every node line it will shard the node stream per-thread; high cardinality.

The named thread (Forge, etc.) found in `data.thread` that appears on Call events is already labelled in the call_trace processor.  Is that sufficient?

If there's also a need for thread on non-Call node lines, happy to dig into it.  The raw log copy has the field available for parsing, but the raw streams aren't uniformly labelled across o-leios and playground so queries on them aren't portable across environments.  To make them portable, it would need to be made available on the enriched/processed stream.  Loki structured metadata could probably do this if we need it.

demo/proto-devnet/config/alloy-modules/tx_centrifuge_enrich.alloy
johnalotoski: Similar to the `type` drop, we can rewrite the dashboards using the `generator` label and simplify to `service` usage only.  Will do!

demo/proto-devnet/config/alloy-modules/tx_centrifuge_enrich.alloy
johnalotoski: Because:
1) The guard comes first; `ns` is not parsed and available yet.
2) Parsing could be moved before the guard but then every log line, not just this module's own, would be parsed only to drop the rest. Keeping the guard first and matching on the already-available routing label avoids that overhead.
3) Matching with ns isn't portable across all three enrichment/process modules -- it can't [in an easily maintainable way] express "keep node" whose ns space isn't a single prefix.

demo/proto-devnet/config/alloy-modules/tx_firehose_enrich.alloy
johnalotoski: Same answer as above.

demo/proto-devnet/config/alloy.template
johnalotoski: It is just a different topology but correct and as intended.  It is also now updated slightly from the drop of the `stream` label in favor of a single `service` label.  It has the benefit of less log write volume for the same useful content.

Example for consideration of the enriched output: 
* Pre-existing topology would query `{service="cardano-node"}` for all cardano-node logs at the cost of ~3x storage
* This new topology would query `{service=~"cardano-node.*"}` for all cardano-node logs without duplication.

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
johnalotoski: Added!

demo/proto-devnet/config/alloy.template
johnalotoski: Commit pushed.

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
johnalotoski: Commit pushed.

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
johnalotoski: Done!

demo/proto-devnet/config/alloy-modules/call_trace_enrich.alloy
johnalotoski: Done.

demo/proto-devnet/config/alloy-modules/leios_voting_enrich.alloy
johnalotoski: Done!

demo/proto-devnet/config/alloy-modules/node_enrich.alloy
johnalotoski: Done.

demo/proto-devnet/config/alloy-modules/tx_centrifuge_enrich.alloy
johnalotoski: Done!

bladyjoker (approved):
Thank you John, looks great!

PR #1038, opened 2026-08-08
trace-verifier: distinct EB content + ToPropose modelling — kill the last mapping-induced false convictions

ramsay-t:
Stacked on #1037. Closes the last two harness-side false-conviction classes so
that live verification reports **only genuine spec/implementation
disagreements** — no violations caused by the mapping failing to read what the
log provides.

Root cause was a single constant: every mapped EB carried synthetic txs `[0]`,
and Defaults hashes an EB to its txs, so **all EBs shared one hash**. That
produced (a) wrong-EB pairings in the VT-Role `find` whenever two EBs
coexisted ("Hashes mismatch, ebHash={0}" — the `{0}` is the collision value),
and (b) permanently-empty `ToPropose`, convicting every own-EB forge
(`Err-EB-Role-premises`).

Fix (mapping only, no parser/type changes): synthetic txs derived from the
EB's on-wire hash string, and the SUT's forges bracketed with `Base₁` steps —
`SubmitTxs (txs eb)` before the `EB-Role` action, `SubmitTxs []` after.

**Field validation** (live producer NITEN on Musashi): before — ~14
`Err-EB-Role` + ~4 hash-mismatch convictions/day; after deployment — zero of
either in the post-deploy window, residual violations exclusively the two
known timing-divergence classes (vote deadline & receipt window — the design
question raised on #1037).

Known honesty gap, documented in the commit: the node logs tx *counts*, so EB
content is verified structurally rather than literally; literal verification
needs closure data from the LeiosDb (observer-mode work).

### DIFF
--- a/leios-trace-verifier/hs-src/app/linear-chain/Main.hs
+++ b/leios-trace-verifier/hs-src/app/linear-chain/Main.hs
@@ -270,11 +270,15 @@ queryChain LeadershipOpts{..} = do
         era
   case result of
     Left err -> die ("local state query failed: " <> show err)
-    -- A stakeless pool (e.g. freshly registered on a young network) has no
-    -- leadership schedule; that just means no winning slots this epoch.
-    Right (Left lerr, stakeDistr) -> do
-      hPutStrLn stderr ("leadership schedule unavailable (" <> show lerr <> "); assuming no winning slots")
+    Right (Left (Api.LeaderErrStakePoolHasNoStake _), stakeDistr) -> do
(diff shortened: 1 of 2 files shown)


ramsay-t:
Superseded: both commits were recreated (with fixes on top) in javierdiaz72/trace-verifier-cardano-api-fixes, and that line — merged with main and re-pinned to the CIP-aligned spec — now lives on ramsay-t/cip-spec-repin, field-validated on Musashi. Closing in favour of a PR from that branch.

PR #1063, opened 2026-08-28
Trace verifier: Fixes to issues found on devnet and testnet

javierdiaz72:
Brings the Linear Leios chain trace-verifier work onto current `main`, which had
meanwhile merged #963 and repinned `leios-spec` to the CIP-aligned spec. That pin bump is an API change, so the merge commit is a port rather than a textual merge.

Two commits are @ramsay-t's, replayed with authorship preserved.

## What changed, by theme

**Made the verifier able to check anything at all.** `ToPropose` is re-established each slot, so `EB-Role`'s first premise is satisfiable; the EB payload is derived injectively from its hash, so distinct EBs don't collapse to one spec-level hash; the in-progress slot is no longer adjudicated at a checkpoint.

**Corrected what raises an obligation.** Voting keys on *announcement* rather than acquisition; a slot is adjudicated against the EB actually voted for; a vote cannot establish its own precondition; only a tip strictly *beyond* an announcer retires its EB; votes for ranking blocks adopted on a fork are resolved rather than dropped.

**Stopped false positives from killing sessions.** An EB is demanded only when the mempool could not have fitted in the ranking block, judged by the pre-drain reading in a slot that forged. An empty leadership schedule is trusted only past the two-epoch stake activation delay. A gap in the slot ticks restarts verification instead of being rejected. EB-role enforcement is gated on Leios being active, and that gate is reported only where it costs something.

**Made it usable on long runs.** The segmentation driver moved into the library so it is testable with a stubbed query and a collecting reporter; segments are re-verified periodically rather than at every tick (the quadratic cost had the verifier starving the node it observes); segments overlap so straddling obligations are still checked.

## The port

`validityCheckTime` is gone from `SpecStructure`, replaced by the `isValidityChecked` predicate. `σc-num`/`σc-den` moved into `Params`, with committee selection now `inVotingCommittee` in `Leios.Config`. A key is registered for every party, without which acquired EB headers fail `isValid` and never reach `EBs'`.

The new spec also changes voting materially: `VT-Role` is no longer satisfiable in exactly one slot. The window opens at `ebSlot + 3·Lhdr` and runs to a deadline at `ebSlot + 3·Lhdr + Lvote`, and a new rule `Roles₃` licenses abstention anywhere inside it. A vote is forced only at the deadline. Consequently `overlapSlots` is now `3·Lhdr + Lvote + Lhdr` = 8, not 4 — carrying only 4 across a boundary would silently drop votes falling due late in the window.

Two of `main`'s changes are deliberately not taken, being superseded here: treating *any* leadership error as an authoritative empty schedule (which reported an epoch-0 devnet forge as violating `EB-Role`), and `CEBAcquired`/`CVoted` setting `curEB`.

## Verification

- 67 examples pass; the executable builds.
- The carry width has a discriminating test: it fails with the old width
  (`not expected: []`) rather than merely asserting a number.
- Byte-identical output to the pre-merge binary over 830 slots of devnet traffic at 100 TX/s — no regression.
- With 50-slot epochs, the diff against the pre-merge binary shows the expected
  difference at five boundaries (segments starting at `boundary − 8` versus
  `boundary − 4`) with no violations from either, so the wider carry adds coverage without false positives.
- Several fixes are corroborated independently by the node's own
  `CNotVoted … chainTipDoesNotAnnounce` reason field.

## Known limitations, and one open question

**The queried leadership schedule can never govern a segment.** `segmentAuthoritative` requires `not spans`, and the boundary carry always carries something, so only the very first segment can ever be authoritative — and only if the node answers for that same epoch. Eligibility therefore always falls back to the log's self-report, which is the weaker check. The `not spans` reasoning is sound in itself (an EB in the carried tail belongs to the previous epoch), but the fix is per-slot epoch attribution rather than relaxing it. Filed as follow-up; it means the cardano-api schedule query is currently inert.

**Votes for the node's own EBs go unverified.** `VT-Role` requires the EB in `EBs'`, which `Slot₁` fills only from received messages, so the spec models no vote for an EB you produced. Measured at 1–3 of every 8 votes, scaling with the node's share of production. Needs a spec decision, not a translator patch.

**Obligations straddling a tick gap are not checked.** A gap restarts verification carrying nothing, since a carried tail would still contain the gap. One devnet case had a node-admitted missed vote (`NotVoted … tooLate`) fall through this and the self-forged gap simultaneously.

**Open question for the node team:** the node reported `tooLate` at slot 425 for an EB announced at 423 — two slots *before* the voting window opens under the `Lhdr = 1` that `Main.hs` hardcodes. Either the prototype's Leios timings differ from `(1, 4, 7)`, or `tooLate` tests something other than the voting deadline. If the timings differ, every window calculation here needs revisiting. The node's config carries only tracer settings and nothing in the log echoes the values.

## Note for reviewers

`main`'s devnet devshell does not currently build — `ouroboros-consensus` fails
`-Werror=deprecations` on `encodeVerKeyDSIGN`, and the derivation is in `origin/main`'s own closure, unrelated to this branch. Running the devnet needs a checkout from before the pin realignment.

### DIFF
--- a/leios-trace-hs/src/ChainEvents.hs
+++ b/leios-trace-hs/src/ChainEvents.hs
@@ -32,7 +32,7 @@ import Data.Aeson.Types (Parser, parseMaybe)
 import qualified Data.ByteString.Lazy.Char8 as BSL8
 import Data.List (mapAccumL)
 import qualified Data.Map.Strict as Map
-import Data.Maybe (catMaybes, mapMaybe)
+import Data.Maybe (mapMaybe, maybeToList)
 import Data.Text (Text)
 import qualified Data.Text as T
 import qualified Data.Text.Read as TR
@@ -78,6 +78,13 @@ data ChainEvent
     --   advances past an announced EB that EB can no longer be certified — the
     --   verifier uses this to close the EB's voting window.
     CChainExtended !Word64
+  | -- | The (min, max) mempool size in bytes observed during one slot, aggregated
+    --   from @Mempool.AddedTx@ / @Mempool.RemoveTxs@ (both report the resulting size
+    --   absolutely, not as a delta). A range rather than a point because the node
+    --   decides from a snapshot taken at an instant the log does not pin down, so a
+    --   single reading would be a guess at what it saw. Emitted immediately before
+    --   the slot tick that closes the slot it describes.
+    CMempoolRange !Word64 !Word64
   deriving (Eq, Show, Generic)
 
 -- | Internal: one parsed line — a directly-emittable event, a linkage fact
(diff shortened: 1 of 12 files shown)


javierdiaz72:
From a Slack conversation with @yveshauser:

```
● Code review(review · 8 findings)
  ⎿  leios-trace-verifier/hs-src/src/LinearLeiosChain.hs
       ●  32 [correctness]     The closeLast=true verifier variant (verifyChainTraceFinalFromSlot), tested in Spec.ChainVerifier, is never wired into the production driver — finishSeg still uses the streaming variant that leaves the last slot unadjudicated.
       ● 295 [correctness]     The same-epoch continuation guard requires toInteger s == l + 1 exactly; a duplicate or non-monotonic CSlot satisfies neither this nor the gap condition (s > l + 1), so it silently falls into the epoch-boundary path.
       ● 351 [correctness]     reportIfLeiosInactive prints the "EB-role enforcement is suppressed" warning without checking segAuthoritative, even though the spec only ever falls back to the activity-gated path when the schedule is NOT authoritative.
       ● 225 [maintainability] gateSuppressesAnObligation is an independent Haskell reimplementation of ebOwed (in the Agda spec), kept in sync only by a comment, not by shared code or a cross-checking test.
       ● 318 [simplification]  The gap branch hand-builds Carry{carryEvents = [], carryStart = toInteger s, carrySpans = False} instead of calling carryAcross overlap (toInteger s) [], which already produces the identical record for an empty seen list.
     leios-trace-verifier/src/LinearLeiosVerifierChain.agda
       ● 500 [correctness]     Accumulator.votedEB is a single Maybe, so a second CVoted in the same slot overwrites the first with no VT-Role or No-VT-Role ever emitted for it.
       ● 175 [correctness]     leaderSlots gates EB-production eligibility on "at or after the first Leios-activity slot seen in this segment's prefix" rather than "Leios currently running," so it cannot re-close if Leios goes idle again later in the segment.
       ● 409 [correctness]     annRB gives a same-slot vote unconditional priority over curEB's superseded status, and nextEB keeps curEB = superseded for another slot whenever a vote landed, so the one-time de-announcement can be deferred as long as votes keep arriving.

● The review of PR #1063 (local diff, HEAD vs origin/main, 11 files / +1779 -204) surfaced 8 findings, most severe first:

  1. LinearLeiosChain.hs:32 — the tested verifyChainTraceFinalFromSlot (closeLast=true) path is never wired into the production driver; a complete node.log can exit "ok" while missing a real violation in the final slot.
  2. LinearLeiosVerifierChain.agda:500 — votedEB is a single Maybe, so two same-slot CVoted events silently clobber each other with no adjudication for the first.
  3. LinearLeiosVerifierChain.agda:175 — leaderSlots's activity gate is monotonic once triggered; it can't re-close, reproducing the false-positive class it was meant to fix if Leios goes idle again later in a segment.
  4. LinearLeiosVerifierChain.agda:409 — a same-slot vote defers curEB's superseded transition indefinitely if votes keep arriving, so de-announcement can be delayed well past the intended single slot.
  5. LinearLeiosChain.hs:295 — the epoch-continuation guard (s == l+1) doesn't handle duplicate/non-monotonic slot ticks (e.g. concatenated overlapping logs), which fall through into the epoch-boundary path and corrupt segment state.
  6. LinearLeiosChain.hs:351 — reportIfLeiosInactive doesn't check segAuthoritative, so it can print a false "enforcement suppressed" diagnostic when the authoritative schedule was actually enforcing normally.
  7. LinearLeiosChain.hs:225 — gateSuppressesAnObligation duplicates the Agda ebOwed logic by hand, kept in sync only by a comment.
  8. LinearLeiosChain.hs:318 — the gap branch hand-builds a Carry record instead of reusing carryAcross, a minor duplication.
```

javierdiaz72:
Fixes for findings 2, 4, 5, 6, and 8 are included in 72cfe0fee8d7141cd5be4df9b893b908c44ef80f. Findings 1, 3, and 7 need a decision:

```
# Follow-ups from the review of the devnet trace-verifier branch

Three of the eight review findings need a decision rather than a patch, so they were
left out of `72cfe0fee`. Each is written to stand alone as an issue.

---

## 1. `verifyChainTraceFinalFromSlot` is production-dead, and the two test suites drive
   different entry points

**What.** The executable only ever calls `verifyChainTraceFromSlot` (the streaming
variant, which leaves the trailing slot unadjudicated). `verifyChainTraceFinalFromSlot`
(`closeLast = true`) exists, is exported, and is used exclusively by
`Spec.ChainVerifier`.

**Correction to the review.** The report implied the production path is untested. It is
tested — `Spec.ChainSegments` drives `runSegmented`, which goes through the streaming
variant. So this is a split between suites, not an untested path: unit fixtures validate
one entry point while the executable runs another, and they differ precisely in whether
the last slot is closed.

**Why it isn't just cosmetic.** For a finite input (`cat node.log | …`) the final slot is
complete and never checked, so a violation there is missed. The driver cannot currently
tell a finite input from a live `tail -F` one.

**Options.**
- Delete `Final` and have `Spec.ChainVerifier` drive the streaming path, accepting that
  fixtures need a trailing tick to be fully adjudicated. Removes the split entirely.
  *Preferred.*
- Keep it and add a `--complete-log` flag the caller sets when the input is finite.
- Keep it and document that the last slot of a finite input is out of scope.

---

## 2. The Leios-activity gate cannot re-close once triggered

**What.** `leaderSlots` takes the earliest Leios-activity slot in the prefix and admits
every leader slot at or after it. If Leios goes idle again later in the same segment, the
gate stays open and EB obligations keep being raised.

**Severity, honestly.** Lower than when the gate was written. An EB is now owed only when
the mempool exceeds the ranking block capacity, so an idle-Leios stretch mostly raises no
demands regardless of the gate. The gate is also recomputed per segment, and slot gaps
keep segments short (~34 slots on the observed devnet), which bounds the exposure further.

**Options.**
- Retire the gate: the mempool rule now covers most of what it was for, and one fewer
  hand-maintained heuristic is worth something. *Preferred, but it is a coverage
  reduction and should be a deliberate call.*
- Gate on a trailing window — Leios active within the last N slots — so it can re-close.
  Needs a defensible N.
- Leave it, and note the limitation.

---

## 3. `gateSuppressesAnObligation` duplicates `ebOwed` by hand

**What.** `LinearLeiosChain.gateSuppressesAnObligation` (Haskell, reporting-only) mirrors
the `ebOwed` decision in `LinearLeiosVerifierChain.agda` (the maximum of the slot's
mempool range where the node forged, the minimum otherwise). They are kept in step by a
comment, nothing more. If `ebOwed` changes, the diagnostic silently diverges from what the
verifier actually did.

**Why it exists.** The reporter runs in the driver, which has no access to the Agda
decision; the Agda side cannot report. Sharing the code across the FFI boundary is not
straightforward.

**Options.**
- A table-driven test asserting the two agree over a set of (range, forged, capacity)
  cases. Cheap, catches divergence, does not remove the duplication. *Preferred.*
- Export `ebOwed` from the Agda and call it from the driver, removing the duplicate
  outright. Cleaner, but widens the FFI surface for a diagnostic.
- Drop the diagnostic and accept less visibility.
```

yveshauser (approved):
The unresolved findings are about the static trace verifier (not the chain trace verifier), code duplication, or a probably exceptional case. Let's document those for now and run more tests against the Musashi testnet.

PR #1094, opened 2026-09-10
Dozen devnet patches

pgrange:
I stumbled upon 2 issues when trying to run the dozen demo with local changes in my cardano-node:
1. a crashing cardano-node would create zombie processes because of how run-node.sh forgot to stop the health check sub-process on exit.
2. run-node.sh would not be able to find cardano-node in its PATH because elevated processes run through sudo,  which drops the environment so adding it as a environment variable.

### DIFF
--- a/demo/dozen-devnet/run-node.sh
+++ b/demo/dozen-devnet/run-node.sh
@@ -16,11 +16,17 @@ export LEIOS_DB_PATH="leios.db"
 # so the waiter can only ever see the socket the node actually created.
 rm -f "node.socket"
 
-# Make socket accessible to non-root (node runs elevated for namespace access)
+# Make socket accessible to non-root (node runs elevated for namespace access).
+# Killed via the EXIT trap below so it doesn't outlive cardano-node: if the
+# node dies before creating node.socket (e.g. a startup crash), this loop
+# would otherwise leak as an orphan (reparented to PID 1) polling forever,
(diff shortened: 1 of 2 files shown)


ch1bo (approved):
Thanks. Do these issues also affect `demo/proto-devnet`?

pgrange:
> Thanks. Do these issues also affect `demo/proto-devnet`?

Didn't check that but I can and I'll let you know.

pgrange:
@ch1bo indeed, the same problems existed for proto-devnet. I took the liberty to add a commit to that PR to fix those, I hope it's ok.

demo/dozen-devnet/run-node.sh
```diff
@@ -47,8 +53,10 @@ if [ -n "${NODE_RTS:-}" ]; then
   RTS_ARGS=(+RTS ${NODE_RTS} -RTS)
 fi
 
-# Run cardano-node
-cardano-node run \
+# Run cardano-node. CARDANO_NODE is passed in as an absolute path by run.sh
+# (sudo/is_elevated drops PATH, same reason ip/tc are invoked by absolute
+# path); fall back to a bare PATH lookup when run without elevation (TC=0).
```
ch1bo, line 58: I'm regularly using PATH to run a different node without issues (without needing this fix).

PR #1096, opened 2026-09-13
Port Leios voting observability to w36, add a diffusion facet

johnalotoski:
w36 folded the per-point vote tally into `LeiosVoteAcquired`, which also
changed that trace from firing per arrival to once per accepted vote. This
retargets the initial draft of the voting alloy module and dashboard from 
https://github.com/input-output-hk/ouroboros-leios/pull/1064 onto it, and adds a diffusion facet covering the part of certification 
the voting facet cannot see.

### Voting facet, ported
- Tally read from `LeiosVoteAcquired`; `rbHash` and `voterId` now sit under
  `vote`. `tally_updates_total` dropped, it duplicated
  `votes_acquired_total`.
- `vote_threshold` added, since the quorum is a protocol parameter rather
  than a constant.
- **Capability lost:** vote diffusion redundancy. With no per-arrival trace
  left, arrivals-over-distinct-votes is not computable. That panel now plots
  tally quantiles against the threshold.

### New: `leios_diffusion_process` + diffusion dashboard
24 metrics over announcement relay, fetch scheduling, body hits and EB
validation. Service refined only on the three kinds it exclusively owns —
`LeiosEbValidated` belongs to both facets and the label can name one owner.
`equivocation` is the only added label, a bool.

### Also
- Restores the call-trace dashboard's uid, which a UI export had changed to
  `calltrace`, breaking `/d/cardano-leios-call-trace`.
- CI check for dashboard uids, ids, datasource references and refresh
  defaults. Verified against the two commits that caused this.
- Observable participation no longer rescans 24h of logs per refresh:
  13s → 2.1s, ~1GB → 19MB.
- A second cache hit rate panel on throughput, counter-based, beside the
  existing LogQL one. They weight differently (per transaction vs per
  validation) and are kept side by side so the difference can be judged
  before either is retired.


demo/proto-devnet/config/dashboards/cardano-leios-voting.json
```diff
@@ -1,444 +1,3522 @@
...
+      {
+        "name": "threshold",
+        "type": "constant",
+        "label": "Certification threshold",
+        "query": "0.75",
+        "current": {
+          "text": "0.75",
+          "value": "0.75"
+        },
+        "hide": 2
```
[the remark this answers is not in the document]
johnalotoski: Fixed with a query variable over the gauge.

demo/proto-devnet/scripts/check-dashboards.sh
```diff
@@ -0,0 +1,169 @@
...
+    for a in $ALLOWED_DS; do [ "$ds" = "$a" ] && found=1 && break; done
+    # ALLOWED_DS entries contain spaces, so re-check the multi-word ones exactly.
+    case "$ds" in
+    "-- Grafana --" | "-- Dashboard --" | "-- Mixed --") found=1 ;;
+    esac
+    if [ "$found" -eq 0 ]; then
+      echo "FAIL $base: datasource uid \"$ds\" is not provisioned here; expected one of: loki, mimir, or a \$variable"
+      fail=$((fail + 1))
+    fi
+  done < <(jq -r '[.. | objects | select(has("datasource")) | .datasource | select(type == "object") | .uid? // empty] | unique | .[]' "$f")
```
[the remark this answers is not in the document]
johnalotoski: Fixed.

demo/proto-devnet/config/alloy-modules/leios_voting_process.alloy
```diff
@@ -37,19 +49,78 @@ declare "leios_voting_process" {
...
+			selector = `{kind=~"LeiosEbValidated|LeiosNotVoted|LeiosVoteAcquired|LeiosVoteScheduled|LeiosVoted"}`
 
 			stage.static_labels {
 				values = {
 					service = "cardano-node/leios-voting",
 				}
 			}
 
+			// Certification margin. LeiosVoteAcquired is emitted once per accepted
+			// vote, not once per arrival -- the same vote reaches us from every peer
```
ch1bo, line 61: @johnalotoski Would you have benefited from an once per arrival trace still here? There are lower level traces of the network stack upon send/recv and the message you would want is `Recv` of `MsgLeiosVotes` if you still need that.

demo/proto-devnet/config/alloy-modules/leios_diffusion_process.alloy
```diff
@@ -0,0 +1,402 @@
...
+declare "leios_diffusion_process" {
+	argument "forward_to" { }
+
+	export "receiver" {
+		value = loki.process.enrich.receiver
+	}
+
+	loki.process "enrich" {
+		stage.match {
+			selector = `{kind=~"LeiosBodyHits|LeiosEbValidated|LeiosFetchDecision|LeiosPeerAnnouncement"}`
```
ch1bo, line 51: This sounds like an alloy pipeline that helps us visualize the data diffusion times of EBs. I'm surprised to no not read `LeiosBlockAcquired` and `LeiosBlockTxsAcquired` which basically signal full download (and processed) of the EB body and all its referenced txs respectively.

ch1bo (commented):
Quick read through and also wanted to give this a spin against the `demo/dozen-devnet`

ch1bo (approved):
Works nicely also for the `dozen-devnet` 👍

demo/proto-devnet/config/alloy-modules/leios_voting_process.alloy
johnalotoski: No, not currently I think.  We've silenced those traces due to volume, and the once per-accepted vote dedup is what makes the tally panels work. If we need it later, a Recv MsgLeiosVotes counter rather than traces would do it.

demo/proto-devnet/config/alloy-modules/leios_diffusion_process.alloy
johnalotoski: You're right, we have a gap there.  We have partial coverage in the voting dashboard for LeiosBlockAcquired via Loki queries, but nothing for LeiosBlockTxsAcquired.  See https://github.com/input-output-hk/ouroboros-leios/pull/1102 for cleaner separation between voting dashboard and diffusion dashboard, and coverage of LeiosBlockAcquired and LeiosBlockTxsAcquired.

